PR #3, opened 2017-01-13
Lint fixes

typpo:
- Enable eslint to run by removing deprecated option
  - Apply existing lint rules in eslintrc (fixes all lints except for console.log warning)


mldangelo:
thanks @typpo!

PR #9, opened 2017-02-06
Update hamburger style

mldangelo:
Make's the hamburger prettier
added a deploy script to make deploying easier. 
refactored links from both menus into one file.


typpo:
:+1:

PR #18, opened 2017-03-25
User Authentication + Private CV

mldangelo:
#### New Features: 
- Adds passport.js with google oath
- Adds private routes. Protects /resume
- moves /resume data to be served by an api 
- adds mongo as a dependency and uses mongo session store

#### Enhancements
- Switches to new react-router v4 API
- Moves webpack express interface into separate file
- Upgrades all dependencies to their latest versions
- Moved `/pages` to `/views` to follow popular conventions 
- Hamburger menu links now change color on hover 
- moved everything into a server folder

#### Bug Fixes: 
- No longer serving entire app directory in production mode

#### Known issues: 
- `app/data/protected` is used by the server and is not monitored by nodemon. Changes to this folder require nodemon to be restarted
- Disabled mocha test for resume because I'm not exactly sure how to authenticate with GoogleOauth in a headless environment 
- No test exists for new login page
- The application structure is continuing to become less organized. Some files are in the routes folder which have very little to do with routes
- PropType definitions for PrivateRoute are most likely not correct
- Login does not remember previous page. It always redirects to /resume
- Google Analytics does not record page hits from 404s


typpo:
👍

PR #233, opened 2020-05-16
Bump Dependencies, utilize hooks API, update text

mldangelo:
This PR introduces a number of important changes described in detail below. This is the first in a series of upcoming PRs that will aim to simplify development and deployment of this website and modernize it to use best practices. 

# Changed 
- Bumps node from v11 to the latest stable release of v12. This should have been done a long time ago, as even-numbered releases have long term support and v11 is no longer actively being maintained. See [here](https://nodejs.org/en/about/releases/). 
- Uses [react-snap](https://github.com/stereobooster/react-snap) to set up paths when using static hosting such as github pages - removes `parseRoutes.js`. This simplifies the deployment process. 
- fixes a number of typos 
- Makes many components functional. Converts analytics and scrolling to top to use [react hooks](https://reactjs.org/docs/hooks-reference.html).
- Bumps all dependencies to latest stable versions. 
 
# Removed
- references to NGINX, as NGINX is not necessary when deploying statically and has confused a lot of people here.


app/components/Template/Analytics.js
```diff
@@ -0,0 +1,24 @@
+import { useEffect } from 'react';
+import { useLocation } from 'react-router-dom';
+import ReactGA from 'react-ga';
+
+if (NODE_ENV === 'production') {
```
will-holley, line 5:
```javascript
IS_PRODUCTION = NODE_ENV === 'production'
```

replace on `:5` and `:13` ...slacking on the reviews...

PR #265, opened 2021-01-20
Alternate email links non-stop

YangZhou0:
Before the change, the email links stop alternating at 'thanks@mldangelo.com'. 
The new change allows the email link to keep updating non-stop. So after 'thanks@mldangelo.com', it rotates back to  'hi@mldangelo.com' and repeat.

### DIFF
--- a/src/components/Contact/EmailLink.js
+++ b/src/components/Contact/EmailLink.js
@@ -45,7 +45,7 @@ const useInterval = (callback, delay) => {
   }, [delay]);
 };
 
-const EmailLink = (props) => {
+const EmailLink = ({ loopMessage }) => {
   const hold = 50; // ticks to wait after message is complete before rendering next message
   const delay = 50; // tick length in mS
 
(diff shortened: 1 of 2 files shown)


mldangelo:
Hi @yangzhou93, thank you for submitting a PR. Not looping was an intentional design choice, but I am supportive of this feature. How would you feel about optionally looping via a prop? We could introduce something like:

```jsx
<EmailLink repeat />
// or 
<EmailLink loop={true} />
```
I would be happy to accept either. Thanks again!

yangzhou93:
> Hi @yangzhou93, thank you for submitting a PR. Not looping was an intentional design choice, but I am supportive of this feature. How would you feel about optionally looping via a prop? We could introduce something like:
> 
> ```js
> <EmailLink repeat />
> // or 
> <EmailLink loop={true} />
> ```
> 
> I would be happy to accept either. Thanks again!

Thank you for the acknowledgment and suggestions. Please see the new PR with the suggested change.

mldangelo:
Thank you so much for working on this. I'm sorry it took so long to land your change.

Issue #272, opened 2021-02-08
Smooth Scroll in Resume Page


mrkazmi333:
If the site can have a smooth scroll when the user clicks on a certain option in this page like Experience/Education/Skills etc. It'll be  add to a great user experience as compared to current one which takes the user to that section in a flash.
I would be really happy to improve this and work on it.

mldangelo:
Hi @mrkazmi333, I will also welcome this PR, as long as the scroll is relatively quick. You may be able to accomplish this with https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior Thank you so much for volunteering to help.

PR #286, opened 2021-05-15
Add smooth scroll support and fixed linebreak errors

rayliuca:
ref #272 

Since I want this feature as well and it's quite simple, I just implanted it myself.

the css entry for smooth scroll is in the src/static/css/base/_page.scss file

linebreaks are fixed by `eslint src --fix` (was CRLF when I cloned the repo), no other files are changed.

[my website](https://rayliu.ca/) is still a work in progress so it might change, but you can see what the change would look like

### DIFF
--- a/src/static/css/base/_page.scss
+++ b/src/static/css/base/_page.scss
@@ -5,7 +5,11 @@
 ///
 
 /* Basic */
-
+	// For smooth scroll support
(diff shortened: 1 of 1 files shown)


mldangelo:
This is a great addition. Thanks so much @rayliuca

PR #296, opened 2022-02-05
Update github-pages.yml

DhruvDoshi:


mldangelo:
Thanks but I would like to keep the wording as is.

PR #870, opened 2026-04-07
fix(homepage): clip hero background to prevent mobile overflow

YassirSellami:
Thanks for catching the mobile overflow issue.

We updated the implementation on this branch after reviewing it locally in browser screenshots.

## Summary
- preserve the original `1200px x 800px` hero glow on desktop
- clip the decorative hero background inside `.hero-bg`
- remove horizontal scrolling on mobile without shrinking the desktop effect

## Validation
- verified at `390px` wide: overflow drops from `405px` on `main` to `0` on this branch
- confirmed the desktop hero styling stays aligned with `main`
- `npm run build` passes

## Screenshots
Captured locally against `main` and this updated PR branch.

### Mobile (`390px` wide)
| Before (`main`) | After (updated PR) |
| --- | --- |
| Before mobile overflow | After mobile fixed |

### Desktop (`1440px` wide)
| Before (`main`) | After (updated PR) |
| --- | --- |
| Before desktop baseline | After desktop preserved |

### DIFF
--- a/app/styles/pages/home.css
+++ b/app/styles/pages/home.css
@@ -166,8 +166,8 @@
   top: 35%;
   left: 50%;
   transform: translate(-50%, -50%);
-  width: 1200px;
-  height: 800px;
+  width: 60%;
+  height: 75%;
   background: radial-gradient(
     ellipse 100% 100% at 50% 50%,
     var(--color-accent-light) 0%,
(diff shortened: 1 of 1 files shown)


mldangelo:
Thanks for catching this and opening the PR. We updated the branch after testing it locally in a browser: the mobile overflow is fixed now, and the implementation keeps the original desktop hero treatment instead of shrinking the background effect globally.

mldangelo:
Thanks @YassirSellami for the fix. I verified the mobile overflow behavior locally across narrow viewport widths, and this cleanly addresses it without affecting the hero content.

PR #883, opened 2026-05-01
fix(home): prevent hero gradient from overflowing the viewport

hsnmnr:
## Summary

The decorative `.hero-gradient` on the homepage is hardcoded at `width: 1200px`. On any viewport narrower than 1200px (which is most laptops, all tablets, all phones), the absolutely-positioned element extends past the viewport edge and grows the document horizontally. The fixed `.site-header` then stretches with the document, pushing the burger menu off-screen on mobile.

This only affects `/` because `.hero-gradient` only exists in `Hero.tsx`.

## Repro

1. Open `/` in a browser
2. Resize the window to <1200px wide (or open on any phone/tablet)
3. Notice horizontal scroll bar at the bottom of the page
4. On a narrow mobile viewport, the burger menu in the header is positioned off-screen to the right — only visible after scrolling horizontally

SCR-20260501-nxvh

## Fix

```diff
- width: 1200px;
+ width: min(1200px, 100vw);
```

- ≥1200px viewport: identical to the original — 1200px focused glow
- <1200px viewport: caps at viewport width, so no horizontal overflow

Used `100vw` rather than `100%` because the parent `.page-main` has `max-width: 800px`, which would shrink the gradient too much (`100%` would resolve to 800px and shrink the glow on every screen size).

## Test plan

- [x] `npm run type-check` clean
- [x] `npm run lint` clean
- [x] `npm run format:check` clean
- [x] `npm test` — 309/309 passing
- [x] `npm run build` clean
- [x] Verify homepage looks identical at ≥1200px viewport
- [x] Verify no horizontal scroll on narrow window (<1200px)
- [x] Verify burger menu visible at right edge of header on mobile

### DIFF
--- a/app/styles/pages/home.css
+++ b/app/styles/pages/home.css
@@ -166,7 +166,7 @@
   top: 35%;
   left: 50%;
   transform: translate(-50%, -50%);
-  width: 1200px;
+  width: min(1200px, 100vw);
   height: 800px;
   background: radial-gradient(
     ellipse 100% 100% at 50% 50%,


hsnmnr:
Closing as a duplicate of #870 — apologies for missing it in my pre-flight search (I was searching different keywords).

For maintainer awareness in case it's useful when reviewing #870: the two fixes have a subtle desktop difference. `<Hero />` lives inside `<main class="page-main">` which has `max-width: 800px`, so `.hero-bg` is also capped at 800px on desktop. PR #870's `overflow: hidden` at `.hero-bg` therefore clips the gradient to 800px width on desktop, shrinking the original 1200px focused glow. The approach in this PR (`width: min(1200px, 100vw)`) preserves the original 1200px on screens ≥1200px while still capping at viewport width on smaller screens. Visually subtle either way, but flagging it in case it matters.

Going with #870 — appreciate the work there.

PR #933, opened 2026-07-28
fix(a11y): announce skill-filter results and clarify proficiency cues

mldangelo:
This fixes three accessibility defects in the résumé skills section and removes presentation data that implied distinctions it did not actually encode.

### Proficiency text had no reliable accessible presentation

Each skill was a bare `<span aria-label="…">`. A plain span maps to the generic role, for which author naming is prohibited, so the numeric proficiency label was not a reliable accessible equivalent. The visual tag still varied in weight and text tone by its 1–5 score; that cue needed real text, not an ARIA label on an element that could not use it.

### Filter state did not explain the result

The category buttons exposed `aria-pressed`, which correctly described which control was active, but filtering hid five of six groups without announcing the resulting count or associating the controls with the changed region.

### Category colour carried no category information

Every generated category received the same `var(--color-accent)` value. That value flowed through the data model, component props, an inline custom property, and a decorative tick, but every tag rendered the same tick. It was unrelated to proficiency and distinguished no categories.

## What changed

- Replaced the discarded numeric ARIA label with real visually hidden text on each skill tag: `deep knowledge`, `working knowledge`, or `familiarity`.
- Added one compact visible legend—`Knowledge  Deep · Working · Familiar`—instead of repeating a rating inside every tag or exposing a numeric axis.
- Renamed the tier classes for what they mean; the top tier is `deep`, not the ambiguous `core` (which could mean importance rather than proficiency).
- Kept the existing data-derived weight/tone hierarchy as a visual reinforcement of text that is now also available non-visually.
- Added one polite, atomic status region whose count is derived from the groups actually rendered and deduplicated by skill title: `Showing 3 of 21 skills in Languages.`
- Associated every filter button with both the result container and status region through `aria-controls` using collision-safe React IDs.
- Preserved the print invariant: inactive groups remain in the DOM behind `hidden`, and print CSS restores every group regardless of the screen filter.
- Removed `Category.color`, the identical accent assignment, the component prop/custom-property chain, the decorative tick, and its now-inert print override.
- Updated focused tests and durable `AGENTS.md` guidance to pin the actual behavior and history.

The accessibility rationale follows [ARIA in HTML's naming-prohibited rule](https://www.w3.org/TR/html-aria/#docconformance) for generic elements and [WAI's `role=status` technique](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA22). `role="status"` supplies polite/atomic semantics; the explicit attributes make the intended full-message announcement unambiguous across implementations.

## Verification

- `npm run format:check` — pass (150 files)
- `npm run lint` — pass (177 files)
- `npm run type-check` — pass
- focused CategoryButton/Skills/SkillTag/data tests — 47/47 pass
- full Vitest suite — 464/464 pass
- `npm run build` — pass (16 static routes; expected offline GitHub-stats fallback)
- `npm run verify-export` — pass (14 pages)
- `git diff --check` — pass
- browser QA at 1280px and 390px — correct pressed state, `3 of 21` status, all controlled IDs resolve, no horizontal overflow, and the revised legend stays readable in light and dark themes

This intentionally avoids a new visible 1–5 self-rating scheme. The UI gives readers one coarse knowledge-depth key, while assistive technology receives the corresponding description with every skill.


mldangelo:

mldangelo (commented):
Verdict: **mergeable, but I would reconsider the visible proficiency vocabulary before landing.**

The underlying PR is necessary and the implementation fixes real accessibility defects. In particular:

- replacing a discarded accessible name on a generic `span` with real text is the right fix;
- the live filter result is derived from the rendered data rather than a hand-maintained count;
- the filter still hides rather than unmounts groups, so print behavior remains intact; and
- deleting the category-colour pipeline is a worthwhile simplification because it conveyed no category-specific information.

I independently ran the targeted Skills/SkillTag/data tests (40/40), the complete suite (57 files, 462 tests), type-check, and a production static export. I also checked `/resume/` at 1200px and 390px. The live status correctly changed to “Showing 3 of 21 skills in Languages.”, only the Languages group remained visible, and there was no horizontal overflow.

My design concern is the repeated visible `CORE` / `WORKING` / `FAMILIAR` annotation on every tag. It materially increases the visual noise of what was previously a quick-to-scan skills section. More importantly, the three words are not a parallel scale: “working” and “familiar” describe proficiency, while “core” more naturally describes importance or centrality. A reader can reasonably interpret “core” as “central to my work,” not “highest proficiency.” That makes the new cue more explicit but not necessarily more precise.

I would either:

1. rename the top tier to a proficiency term parallel to the other two (for example, `advanced`, if the data can honestly support that claim);
2. show one compact legend and preserve the quieter tags; or
3. retain the tier as screen-reader text while using a less repetitive visible treatment.

A small implementation nit: `role="status"` already gives the region implicit polite live behavior, so pairing it with an explicit `aria-live="polite"` is redundant. It is harmless, but one mechanism would be clearer.

This is not a functional blocker—the accessibility behavior, print invariant, responsive layout, tests, and export all held up—but the terminology becomes public résumé copy, so it deserves a deliberate editorial decision rather than being treated as just an accessibility implementation detail.

mldangelo:
Copilot's overview raised no actionable defect on this PR, and the substantive items it did flag elsewhere in the batch are tracked on their own PRs.

For traceability, this PR's three changes were each verified against the source before landing rather than taken from the brief:
- The per-category `color` field really did assign `var(--color-accent)` to every one of 27 tags, so `--tag-color` and the per-tag tick carried zero information. Removed with its test.
- `aria-pressed` reports the control, not the result, so the filter now announces the outcome through a live region.
- `SkillTag`'s `aria-label` sat on a bare `<span>`, which maps to role `generic` and has its accessible name discarded — so "proficiency 5 of 5" was announced to nobody. Replaced with real text.

AGENTS.md's constraint that `Skills` keeps every group in the DOM behind `hidden` so `print.css` can restore them is preserved; the built HTML still contains all groups, not just the active one.

mldangelo:

src/components/Resume/Skills/SkillTag.tsx
```diff
@@ -1,41 +1,39 @@
...
-  const categoryColor = categories.find((cat) =>
-    category.includes(cat.name),
-  )?.color;
+/**
+ * Three coarse tiers derived from the 1–5 self-score.
+ *
+ * The score is a self-assessment, so it is reported at the precision it
+ * actually has: a tier, not a plotted value with an axis. The tier is rendered
+ * as real text rather than encoded in the tag's colour, which is what the tag
+ * used to do — a colour-only cue fails WCAG 1.4.1, and the `aria-label` it was
```
mldangelo: The old tag did not encode proficiency in color. `categoryColor` was category metadata (and happened to be identical for every category); proficiency was encoded by the `skill-tag--lg/md/sm` size classes. Real text is still the right accessibility fix, but please correct this history and frame the defect as ambiguous size-only information plus a discarded generic-span accessible name—not a WCAG 1.4.1 color-only competency cue.

mldangelo (commented):
Fresh exact-head re-review at `79b71524a7acbf723f540a7a5a4856448454477f`.

**Verdict: functionally mergeable; one factual code-comment correction and one public-copy decision remain.** This PR is necessary. The old generic-span `aria-label` did not give the proficiency text a reliable accessible presentation, `aria-pressed` described the filter controls but not their result, and the category-color prop chain produced the same accent value for every category. Real text, a status region derived from rendered groups, and removal of the no-information color field are concrete improvements.

The live-region arithmetic is careful: duplicate skills across categories are deduplicated for “all,” category counts reflect the visible group, inactive groups remain mounted behind `hidden`, and print restores all groups. Responsive and printed output remain stable.

There is one factual error in the new explanatory comments. The old implementation did **not** encode competency in color. It encoded competency through `skill-tag--lg/md/sm` size classes; the category color was meant to encode category, but all categories happened to receive the same value and therefore encoded nothing. Adding text still fixes the right problem—size-only proficiency is ambiguous and the discarded ARIA name did not rescue it—but the comments at `SkillTag.tsx:15-18` and `skills.ts:128-133` currently teach the wrong history and invoke WCAG 1.4.1 for the wrong cue. Please correct that rationale before it becomes durable AGENTS guidance.

The remaining design point is editorial rather than functional. Repeating `CORE`, `WORKING`, or `FAMILIAR` on every tag adds substantial visual density, and the terms are not a parallel proficiency scale: “core” usually means central/important, while “working” and “familiar” describe ability. The fresh desktop/mobile rendering is usable, but the labels make the résumé more self-rating-heavy. I would prefer either a parallel scale with defensible words, a single legend, or visually quieter/screen-reader-only tier text. If the repeated labels are intentional, record that as a deliberate content choice rather than an accessibility necessity.

Independent exact-head QA:
- `format:check`, Biome (177 files), TypeScript, and `git diff --check`: pass
- Vitest: 57 files / 462 tests pass
- desktop and 390px browser review: no horizontal overflow; all filters/tags remain readable
- 5-page Letter PDF rendered and inspected; every group prints despite screen filtering, and the added tier labels fit without clipping

No source edit is needed beyond correcting the explanatory history; the implementation itself fixes real accessibility and dead-data problems.

mldangelo:

mldangelo (commented):
Final exact-head re-review at 01a952a.

Disposition: no remaining blocker. The skills filter now communicates both control state and result: the visible proficiency legend remains available, each tier has an accessible textual equivalent, and filtering announces the rendered count without changing what the printed résumé contains. The dead per-category color field is gone, eliminating decoration that implied distinctions it did not actually encode.

I rechecked keyboard/screen-reader semantics, filtering versus print behavior, wording, and data/render consistency. Formatting, lint, type checking, 463 tests, build/export checks, and browser QA pass. This is a real accessibility improvement without introducing a new self-rating scheme.

mldangelo (commented):
Sequential improvement pass on exact head `4bedd66` (COMMENT, not APPROVE, because this is our own PR).

This pass found two remaining quality problems and one small dead-code remnant.

First, the previous remediation still described the old proficiency cue as “size and weight,” but the original `skill-tag--lg/md/sm` rules never changed font size; they changed font weight and text tone. That history is now corrected in the component, CSS, and durable guidance. The per-category colour is also described separately and accurately: it was meant to distinguish categories, but every category received the same accent value, so it encoded neither category nor proficiency.

Second, `CORE / WORKING / FAMILIAR` was not a coherent public proficiency scale. “Core” normally means central or important, while the other two words describe knowledge. The visible legend is now `Knowledge  Deep · Working · Familiar`, and each tag carries the more natural hidden equivalent `deep knowledge`, `working knowledge`, or `familiarity`. The legend remains a single compact key rather than repeating ratings on every tag or exposing a numeric 1–5 axis.

The filter contract is also stronger: each button now uses `aria-controls` to identify both the changed skill-group container and its status region; React `useId()` avoids duplicate IDs; and the polite status is explicitly atomic so the full derived sentence is announced. This follows [ARIA in HTML's naming-prohibited guidance](https://www.w3.org/TR/html-aria/#docconformance) for the old generic-span label and [WAI's status-message technique](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA22). The obsolete print override for the deleted tag tick is now gone as well.

Independent browser verification at 1280px and 390px confirmed:
- `Languages` becomes the sole pressed filter;
- the status changes to `Showing 3 of 21 skills in Languages.`;
- five inactive groups remain in the DOM behind `hidden`;
- every `aria-controls` target resolves;
- the revised legend remains compact and readable;
- no horizontal overflow appears.

Verification on this head:
- focused CategoryButton/Skills/SkillTag/data tests: 47/47 pass
- format check: pass (150 files)
- Biome: pass (177 files)
- TypeScript: pass
- full Vitest suite: 464/464 pass
- static-export verification: pass, 14 pages
- `git diff --check`: pass
- GitHub CI: code quality, tests, macOS, Windows, Node 22, and Node 26 builds pass

No remaining actionable issue from this pass. I did not merge, close, or approve the PR.

PR #934, opened 2026-07-28
fix(resume): derive an honest career timeline and tenure

mldangelo:
## Why

The resume source contains overlapping jobs, internships, and a long-running part-time role. Rendering raw source order had already allowed the timeline to drift, while sorting only by start date made the public chronology worse by placing short earlier roles above an eight-year CTO role that ended in 2022.

The page also hand-maintained a `15+ years` claim and made readers calculate every role's duration themselves.

## What changed

- Added `src/lib/career.ts` as the single home for timeline ordering, role durations, and elapsed career span.
- Sorts completed roles by end-date recency, places an ongoing full-time role first, and places an open-ended part-time role by its start date so it does not permanently outrank later primary work.
- Marks Skeptical Investments explicitly as `commitment: 'part-time'`. `Position` is documented as a JSON Resume-derived work item with this local extension.
- Keeps source-array order non-load-bearing and validates every date as a real `YYYY-MM-DD` calendar date; impossible values such as `2026-02-31` now fail tests instead of being normalized silently by JavaScript.
- Derives visual hierarchy from the data while excluding part-time side roles from the lead tier, including when a future side role has the newest start date.
- Replaces the stronger active-tenure implication with the precise phrase `career spanning N+ years`; `careerSpanYears()` is explicitly elapsed time from the earliest listed role, not a sum of non-overlapping work intervals.
- Reads `Date.now()` once for the page and threads that shared instant through the headline and all ongoing-role durations.
- Displays compact duration text in the date gutter and exposes an unabbreviated `Duration: ...` label to assistive technology.
- Uses existing design tokens and dependencies; no new package or print-specific styling was added.

## Current rendered order

1. OpenAI
2. Promptfoo
3. Smile ID
4. Arthena
5. Skeptical Investments
6. Matroid
7. Planet
8. Planetary Resources
9. Facebook
10. UB Nanosatellite Program

This reflects recency of involvement without letting an enduring side role dominate the main career narrative.

## Verification

- `npm run format:check` — pass
- `npm run lint` — 180 files, pass
- `npm run type-check` — pass
- `npm test` — 59 files / 491 tests, pass
- `npm run build` — static production build, 16 routes, pass
- `npm run verify-export` — 14 exported pages, pass
- `git diff --check` — pass
- Browser QA at 1280px and 390px — expected chronology, durations, full accessible labels, and no horizontal overflow

The build's GitHub-stat request used the existing offline fallback because the sandbox could not resolve `api.github.com`; the static build and export both completed successfully.


mldangelo:

src/components/Resume/Experience/Job.tsx
```diff
@@ -37,6 +51,7 @@ export default function Job({ data, tier = 'primary' }: JobProps) {
         ) : (
           <span className="daterange-present">Present</span>
         )}
+        <span className="daterange-duration">{duration}</span>
```
mldangelo: This duration is visually understandable because CSS gives it a new line, but it has no semantic label. In the accessibility tree it is just a generic `4 mo` / `1 yr 8 mo` immediately after the date range. Please give it a complete accessible form—e.g. an sr-only `Duration:` plus unabbreviated text, or an `aria-label="Duration: 4 months"` on the visible abbreviation (with care not to duplicate the text). That preserves the compact gutter without asking non-visual readers to infer what this number means.

mldangelo (commented):
Verdict: **the chronology fix is useful and the arithmetic is correct, but I recommend fixing the tenure label before merge and reconsidering whether every role needs a second date-derived line.**

I independently verified the rendered order and current calculations. The exported résumé is now ordered OpenAI → Promptfoo → Smile ID → Skeptical Investments → Matroid → Planet → Arthena → Planetary Resources → Facebook → UB Nanosatellite Program, and the displayed tenures match the helper output. I ran the targeted career/resume tests (61/61), the full suite (59 files, 488 tests), type-check, and a production static export; all passed. At 1200px the page had no overflow.

The primary issue is semantic, not arithmetic. In the accessibility tree, each date range is followed by an unlabeled generic value such as `4 mo` or `1 yr 8 mo`. The visual line break implies “duration,” but a screen reader encounters something equivalent to “March 2026, Present, 4 mo” with no label explaining the last fragment. The abbreviations also make the public résumé copy terser than it needs to be. The inline comment below suggests a compact visual treatment with a complete accessible label.

There is also a design tradeoff worth making deliberately. The durations duplicate information already present in the start/end dates and add density to the smallest, quietest part of the career spine. Sorting fixes a genuine comprehension problem; showing duration for all ten positions is less clearly necessary. I would consider retaining the derived headline and ordering while dropping the per-role line, or at least reducing its visual prominence after giving it a proper accessible name.

Finally, “15+ years building products” is calculated as elapsed calendar span from the earliest student-program start to now. It is not a union or sum of employment months, and overlapping roles do not change it. That is a defensible résumé convention, but the helper name and PR prose should be explicit that this is career span rather than accumulated tenure. If that wording is kept, future readers should not infer that it measures fifteen non-overlapping years of full-time product work.

mldangelo:
Implemented in #964 (on the integration branch, which carries this work).

- **`Date.now()` read twice** — real, and worth being precise about the mechanism: there is no concurrency, just two sequential reads microseconds apart, and they only diverge if they straddle the anniversary of the earliest role (2011-06-01), the single instant where `totalExperienceYears` steps. A once-a-year unreproducible red build is still worth removing, and freezing is nearly free. The suite now uses `vi.setSystemTime`, unconditionally in `beforeEach` because `vitest.config.ts` sets `sequence.shuffle`. A new case renders five years apart and requires the span to move by exactly five, so the freeze is pinned to actually reach the page.
- **`Job` JSDoc** — the comment claimed the value is never read from the clock here, which the `= Date.now()` default made false. Fixed the *code*, not the comment: `now` is now required, matching `positionDuration` in `career.ts`, which takes it as a required argument for the same reason.
- **`Experience` JSDoc** — fixed the comment instead, because the property it claims is actually true on the default path: the default parameter evaluates once per `Experience` render and the instant is threaded to every `Job`. Different fix for a different problem.

mldangelo:

src/lib/career.ts
```diff
@@ -0,0 +1,133 @@
...
+ * `src/components/Resume/Experience.tsx`, where UTC midnight reads as the
+ * previous day west of Greenwich. `src/data/__tests__/work.test.ts` pins the
+ * date format the comparison depends on.
+ *
+ * Returns a new array; the input is not mutated.
+ */
+export function sortPositions(positions: Position[]): Position[] {
+  return [...positions].sort(
+    (a, b) =>
+      b.startDate.localeCompare(a.startDate) ||
```
mldangelo: **Chronology regression:** sorting primarily by start date places Matroid (2015–2016) and Planet (2014–2015) above Arthena (2014–2022). A résumé reader expects the 2022-ending CTO role to be more recent than those short earlier roles. Sort completed roles by recency/end date, or add an explicit data-level display key for overlaps/side roles; start date alone cannot encode the intended narrative.

app/resume/page.tsx
```diff
@@ -22,16 +23,26 @@ export const metadata: Metadata = createPageMetadata({
...
           <h1 className="resume-title">Resume</h1>
+          {/* The span is counted from the earliest role in `work`, never
+              typed — see `totalExperienceYears`. A hand-written "15+" is
+              correct on the day it is written and wrong from then on. */}
           <p className="resume-summary">
-            Engineering leader with 15+ years building products across AI,
-            security, and infrastructure. I&apos;m currently a Member of the
-            Technical Staff at OpenAI, working on Promptfoo and Codex Security.
-            I help secure AI systems and use AI to find software
+            Engineering leader with {yearsOfExperience}+ years building products
```
mldangelo: This number is elapsed span since the earliest student-program start, not time spent building products. The data includes a May 2012–June 2013 gap and overlapping roles; the union of active intervals is about 13.7 years today while this renders 15+. Either label it as career span/“since my first role,” or compute non-overlapping active tenure before making the stronger “years building products” claim.

mldangelo (commented):
Fresh exact-head re-review at `04970ba8f9dfa37b6b41d6612c8a0e2ac5ec5cd8`.

**Verdict: changes needed before merge.** Deriving date formatting and tenure helpers is reasonable, and the arithmetic tests are thorough, but the central “chronology fix” currently makes the public résumé chronology worse.

The comparator sorts primarily by **start** date. In the rendered/exported résumé that produces:

`… Skeptical (2017–present) → Matroid (2015–2016) → Planet (2014–2015) → Arthena (2014–2022) …`

An eight-year CTO role ending in 2022 is therefore filed below a nine-month 2015–2016 co-founder role and a 2014–2015 internship. This is not reverse chronology in the way a résumé reader understands it; it prioritizes when overlapping roles began over how recently they were active. The five-page print rendering makes the regression especially obvious: Matroid and Planet occupy page 2, while Arthena begins page 3. The old source order placed Arthena above both.

Overlapping/full-time/side roles mean there may not be one honest comparator derivable from `startDate` and `endDate` alone. Either sort completed roles by recency/end date and decide explicitly how ongoing side roles behave, or add a data-level display-order/sort key so editorial intent is explicit and validated. “A new role can be appended anywhere” is not worth a misleading career narrative.

Two previously noted issues also remain on this unchanged head:

1. Each date range is followed in the accessibility tree by an unlabeled generic value such as `4 mo` or `1 yr 8 mo`. The CSS line break does not provide the missing semantic “Duration:” label. Fix the accessible name or remove the redundant per-role line.
2. `totalExperienceYears` reports elapsed calendar span from the June 2011 student-program start, not years actually building products. The source intervals contain a May 2012–June 2013 gap and many overlaps; their union is about 13.7 years at this head’s review date, while the headline says 15+. If elapsed span is the desired convention, the copy must say that rather than implying 15 non-overlapping years of product work. Otherwise compute the union of active intervals. Automatically increasing the claim every June makes it reproducible, not accurate.

The response points to JSDoc/test fixes on #964, but #934 itself has not moved, and those off-branch changes do not address the chronology or public-copy issues above.

Independent exact-head QA:
- `format:check`, Biome (180 files), TypeScript, and `git diff --check`: pass
- Vitest: 59 files / 488 tests pass
- desktop semantic/browser review confirms all durations and the exact rendered order
- 5-page Letter PDF rendered and inspected page by page; no clipping, but the incorrect Matroid/Planet-before-Arthena hierarchy is visible in print

The PR addresses a real maintenance problem, but its current ordering rule and headline definition are not accurate enough for a public résumé.

mldangelo:

mldangelo (commented):
Final exact-head re-review at 4639b73.

Disposition: no remaining blocker. The chronology is now honest for overlapping and part-time roles: open-ended work is ordered by when it began, completed work by recency, Skeptical is identified as part-time, and the headline says career span rather than overstating active tenure. Role duration has a complete accessible label instead of exposing only an unexplained abbreviation.

The clock-sensitive tests use one deterministic instant and the component comments match their Date.now fallback behavior. All six prior review threads are addressed. Formatting, lint, type checking, 492 tests, production build/export, browser inspection, and résumé PDF QA pass.

mldangelo (commented):
Fresh exact-head re-review at `d75089b3528bd12d18b7176efeb5ce4f760290a2`.

**Disposition: no remaining blocker.** This PR is necessary: the source array had already drifted, the previous start-date-only comparator produced a misleading public résumé chronology for overlapping roles, the headline's elapsed span was maintained as prose, and each role's duration was left to the reader to calculate. The current implementation addresses those problems without adding a dependency or making source order load-bearing.

This pass found and fixed three additional issues:

1. **A future part-time role could become the visual lead.** `sortPositions()` correctly prevents an enduring side role from dominating chronological order, but `tierFor()` independently selected the latest non-student start date. A newly added part-time role with a newer start could therefore receive `jobs-container--lead` even though the data model explicitly calls it a side engagement. Part-time roles are now excluded from lead candidacy, with a regression case that gives a 2027 side role `primary` and keeps the 2026 primary job `lead`.
2. **The data interface overstated schema conformance.** `Position` claimed to conform to JSON Resume while carrying a local `commitment` field. Its comment now accurately says the type is based on the JSON Resume work item with a local extension.
3. **The date invariant accepted impossible calendar dates.** A `YYYY-MM-DD` regex plus `new Date()` validity checks accepts values such as `2026-02-31` because JavaScript normalizes them into March. The data test now round-trips each UTC calendar date and rejects normalization while preserving the string-comparison invariant used by the comparator.

The broader design now holds together:

- completed roles sort by end-date recency, the ongoing primary role leads, and the ongoing part-time role is placed by its start;
- the rendered order is OpenAI → Promptfoo → Smile ID → Arthena → Skeptical → Matroid → Planet → Planetary Resources → Facebook → UB Nanosatellite;
- `careerSpanYears()` and the public phrase “career spanning” accurately describe elapsed span rather than accumulated active tenure;
- one shared instant drives the headline and every open-ended duration;
- compact visual durations are hidden from assistive technology and paired with complete `Duration: …` text;
- hierarchy, dates, and current-role signaling continue to use the existing design system, and desktop/mobile inspection found no overflow or layout regression.

Verification on the exact pushed head:

- `format:check`, Biome (180 files), TypeScript, and `git diff --check`: pass
- Vitest: 59 files / 491 tests pass
- browser QA at 1280px and 390px: expected order, duration labels, status, and no horizontal overflow
- hosted GitHub checks: quality, tests, macOS Node 26, Ubuntu Node 22/26, Windows Node 26, and aggregate CI all pass
- all six earlier review threads remain resolved and outdated

The final hardening changes do not alter the current visible ordering or layout; they make the intended behavior durable when future roles and dates are added.

PR #935, opened 2026-07-28
chore(ci): measure the static export and enforce size budgets

mldangelo:
## Why

`@next/bundle-analyzer` reports the JavaScript module graph, but JavaScript is only about 27% of this static export. HTML, RSC prefetch payloads, images, fonts, CSS, and metadata all become files GitHub Pages serves, and none previously had a reproducible size gate.

A fresh local production export on this head contains 186 files / 3,782,802 uncompressed bytes:

| Subsystem | Files | Bytes | Share |
| --- | ---: | ---: | ---: |
| JavaScript | 33 | 1,009,414 | 26.7% |
| RSC prefetch payloads | 87 | 930,607 | 24.6% |
| Images | 39 | 904,461 | 23.9% |
| HTML documents | 14 | 647,746 | 17.1% |
| Fonts | 4 | 204,352 | 5.4% |
| CSS | 1 | 74,950 | 2.0% |
| Feeds and metadata | 8 | 11,272 | 0.3% |

## What changed

- `scripts/measure-export.mjs` walks `out/`, groups every file, reports the largest files and heaviest route bootstrap sets, measures repeated inline SVG markup, and enforces the committed policy.
- `scripts/budget.json` defines seven positive whole-byte limits with documented headroom. Normal runs require every known metric and reject missing, unknown, malformed, fractional, zero, or negative values; only `--update-budget` can bootstrap a complete policy.
- `.github/workflows/node.js.yml` runs the gate after export verification on all four build legs.
- `scripts/lib/html.mjs` and `scripts/lib/site.mjs` centralize the HTML and base-path/export helpers formerly embedded in `verify-export.mjs`, including repository-site forks under `/repo/`.
- `npm run --silent measure-export -- --json -` keeps stdout valid JSON and sends the human report to stderr. Empty JSON destinations fail loudly.
- JSON reports are rejected inside `out/`, where they would mutate the artifact after it was measured, and cannot overwrite `scripts/budget.json`.
- A missing same-origin stylesheet, script, or preload now fails with route/reference context instead of being omitted and making the route appear lighter.
- No dependency was added.

## Budget semantics

| Metric | Headroom | Purpose |
| --- | ---: | --- |
| `totalBytes` | 20% | catches growth anywhere in the artifact |
| `javascriptBytes` | 15% | dependency/runtime growth |
| `cssBytes` | 15% | design-system growth |
| `fontBytes` | 15% | family/axis changes |
| `maxFileBytes` | 20% | one unusually large asset |
| `maxRouteBootstrapBytes` | 15% | document plus directly linked same-origin styles, scripts, and preloads |
| `repeatedInlineSvgBytes` | 20% | full inline SVG markup repeated beyond its first occurrence |

The route metric is deliberately called a **bootstrap set**, not “first load”: it does not attribute ordinary images, media, `srcset` candidates, or CSS `url()` dependencies to a route. Those bytes remain covered by subsystem reporting, `maxFileBytes`, and `totalBytes`.

Likewise, repeated inline SVG bytes are an upper bound on a deduplication opportunity, not exact sprite savings. SVG reuse still requires a definition and a `<use>` reference at each instance ([SVG 2 `use` element](https://www.w3.org/TR/SVG2/struct.html#UseElement)).

## Deliberate scope

There is no on-page size panel. Writing a report into the export changes the artifact being measured and would require a second build pass to reach a stable answer. This stays a post-build CI/developer tool; reports must be written outside `out/`.

HTML, RSC, and images are reported but not gated as separate aggregate categories because ordinary writing changes them. Material growth still trips the total or largest-file limit.

## Verification

- `npm run format:check` — pass
- `npm run lint` — 182 files, pass
- `npm run type-check` — pass
- `npm run og:check` — pass
- `npm test` — 58 files / 482 tests, pass
- `npm run test:coverage` — pass; 87.33% statements / 84.26% branches
- focused measurer suite — 31 tests, pass
- `npm run build` — production static build, 16 routes, pass
- `npm run verify-export` — 14 pages, pass
- `npm run measure-export` — 186 files / 3,782,802 bytes; all seven budgets pass (82.7%–86.9% used)
- `git diff --check` — pass

The local build used the existing GitHub-stat fallback because the sandbox could not resolve `api.github.com`; the build, exported-site integrity check, and size gate all completed successfully.


mldangelo:

scripts/measure-export.mjs
```diff
@@ -0,0 +1,562 @@
...
+/**
+ * What a cold visit to one route downloads: the document plus every exported
+ * file it asks for before first paint. Resolving those references is why this
+ * shares `exportLayout` with the integrity gate — under a repository-site base
+ * path the markup says `/repo/_next/...` while the file is at `out/_next/...`,
+ * and a hand-rolled prefix strip is how that breaks in forks only.
+ */
+const FIRST_LOAD_RELS = new Set(['stylesheet', 'preload', 'modulepreload']);
+
+function referencedFiles(html, route) {
```
mldangelo: This reference set is not equivalent to everything a cold route downloads. It counts selected `<link>` relations and `<script src>`, but misses ordinary eager images/`srcset`, media/posters, and assets pulled transitively through CSS `url()`. As a result, `maxRouteFirstLoadBytes` can stay green when an actual route gains a large non-preloaded asset. Please either expand the traversal or rename the metric and its budget policy to state exactly this narrower set.

mldangelo (commented):
Verdict: **changes needed before merge.** Measuring the complete export is worthwhile, and the report already exposed useful facts that the existing bundle analyzer misses. However, one of the seven gated metrics is currently named and documented more strongly than the implementation supports.

I independently ran the targeted script tests (29/29), the complete suite (58 files, 463 tests), type-check, production build, `verify-export`, and `measure-export`. The export verified 14 pages and measured 186 files / 3,782,485 bytes; all seven budgets were within limits (roughly 82.7%–86.9% used). The shared base-path extraction appears behavior-preserving.

The important problem is `maxRouteFirstLoadBytes`. The PR describes it as “what a cold visit actually downloads,” but `referencedFiles()` only follows a narrow set of direct HTML references: scripts plus stylesheet/preload/modulepreload links. A normal eager ``, responsive `<source srcset>`, CSS `url(...)` dependency, video/poster, or other document-loaded asset is absent unless separately preloaded. A future unoptimized hero image could therefore make an actual cold route much heavier while this route budget remains green. `totalBytes` may eventually catch it, but that does not make the route metric accurate.

Please either traverse the relevant document and CSS dependency graph enough to support the “cold visit” claim, or rename/document the metric narrowly (for example, direct script/style/font references in the HTML) and avoid presenting it as first-load bytes. A budget is policy, so a misleadingly authoritative name is worse than a smaller, precise metric.

Two smaller interface problems:

- `npm run measure-export -- --json -` prints the human-readable report before JSON. That means stdout is not actually a clean machine-readable JSON stream despite `-` conventionally meaning stdout. Send the human report to stderr in that mode, or suppress it.
- `--json=` silently behaves as though no JSON option was supplied. Invalid option values should fail loudly; otherwise a caller can believe it wrote data when it did not.

Finally, 1,154 added lines is substantial machinery for seven counters in a 3.8 MB personal static site. The measurement is useful, but I would take one simplification pass before making this a long-term CI surface—especially around the bespoke HTML parsing and update-budget CLI. The maintenance cost should be proportional to the regression it prevents.

The inline comment marks the source of the first-load mismatch.

mldangelo:
Both implemented — the second in #965, and the first turned out to be worse than reported.

- **Non-numeric budget silently ungating** — fixed, and the owner review of the integration PR then found the same hole one level deeper: `{}` and `{ "_policy": [] }` pass root validation, every metric gets an undefined limit, and the script reports `0 budget(s) within limits` with status 0. A normal run now requires at least one known metric; `--update-budget` can still bootstrap from empty. Tracked in #971.
- **`--json -` not valid JSON on stdout** — confirmed and fixed. Worth flagging that my first verification of this was wrong: I ran it with `--silent`, which hides npm's two-line banner, so it appeared to pass. The documented `npm run measure-export -- --json -` genuinely fails — `jq` exits 5 and the producer gets `EPIPE`. The documented form and its test now match what a user is actually told to run.

mldangelo:

scripts/measure-export.mjs
```diff
@@ -0,0 +1,562 @@
...
+const budget = readBudget();
+
+const checks = BUDGET_METRICS.map((metric) => {
+  const actual = metric.read(report);
+  const limit = budget[metric.id];
+  return {
+    id: metric.id,
+    label: metric.label,
+    actual,
+    limit: typeof limit === 'number' ? limit : undefined,
```
mldangelo: Treating every absent or non-numeric value as `undefined` silently turns a required CI budget into an advisory report. `{}`, `{ "_policy": [] }`, or string-valued known keys all pass this branch and line 395 marks the metric OK. Require the committed policy metrics to be present, finite, positive numbers on normal runs; reserve an empty budget only for `--update-budget` bootstrap.

scripts/measure-export.mjs
```diff
@@ -0,0 +1,562 @@
...
+    if (arg === '--update-budget') {
+      options.updateBudget = true;
+    } else if (arg === '--json') {
+      options.json = argv[index + 1];
+      index += 1;
+      if (options.json === undefined || options.json.startsWith('--')) {
+        die('--json needs a file path (use `-` for stdout)');
+      }
+    } else if (arg.startsWith('--json=')) {
+      options.json = arg.slice('--json='.length);
```
mldangelo, line 69: `--json=` produces an empty string here, then the script exits 0 without writing a report because the later `if (options.json)` treats it as false. Reject an empty inline value exactly as the separate `--json` form rejects a missing path; I reproduced exit 0/no output on this exact head.

mldangelo (commented):
Fresh exact-head re-review at `2e6c911746c097f5b56a09b6c76480cf4d9360d5`.

**Verdict: changes needed before this becomes a CI policy surface.** Measuring the complete static export is useful: the report correctly shows that HTML/RSC/images are a large part of the 3.78 MB artifact, and the shared base-path/export helpers appear behavior-preserving. But the current gate can silently disable itself, one named metric does not measure what its name promises, and two documented CLI modes still fail.

The response says the JSON and budget-validation problems were fixed in #965/#971. #935 itself has not moved; all of the defects remain on this exact head:

1. **The entire budget can disappear while CI stays green.** `readBudget()` rejects unknown keys but does not require any known key and does not validate known values. `{}` or `{"_policy":[]}` yields seven `undefined` limits; line 395 treats every one as OK and the command exits 0 with `0 budget(s) within limits`. A string value silently ungates only that metric. This defeats the central purpose of a required CI budget file. A normal run should require every policy metric (or at minimum a documented required set) to be present, finite, positive numbers; only `--update-budget` should be allowed to bootstrap.
2. **“First load” is materially incomplete.** `referencedFiles()` follows scripts plus stylesheet/preload/modulepreload links. It misses ordinary eager `img src/srcset`, `picture/source`, media/posters, and transitive CSS `url()` assets. A route can gain a large hero image while `maxRouteFirstLoadBytes` remains unchanged. Traverse the actual dependency set or rename the metric everywhere to the narrower “direct HTML script/style/preload bytes.”
3. **`--json -` is not machine-readable.** I reproduced the documented npm form exactly: `npm run measure-export -- --json - | jq empty` exits 5 with a parse error because npm’s banner and the human table precede the JSON. Even direct Node invocation mixes the table with JSON. Suppress/redirect human output when stdout is the JSON destination.
4. **`--json=` silently succeeds.** I ran `node scripts/measure-export.mjs --json=`; it exited 0 and wrote no report. Empty values should fail at argument parsing, just like `--json` without a following path.

The change is also 1,154 added lines for seven counters on a small personal static site. After correcting the semantics, please take a simplification pass. The durable value is a precise, hard-to-bypass gate; bespoke reporting and inaccurate route-graph machinery are liabilities if they obscure that core.

Independent exact-head QA:
- `format:check`, Biome (182 files), TypeScript, and `git diff --check`: pass
- Vitest: 58 files / 463 tests pass
- production static build and `verify-export`: pass, 14 pages
- `measure-export`: 186 files / 3,782,634 bytes; all seven committed limits currently pass (82.7%–86.9% used)
- fault checks: `--json - | jq empty` fails with jq exit 5; `--json=` exits 0 without producing a report
- shared site/base-path helper diff reviewed; no behavior regression found in the current export

This is a worthwhile capability once its policy cannot silently become advisory and its metric/CLI names match observable behavior.

mldangelo:

mldangelo (commented):
Final exact-head re-review at 97eb0c6.

Disposition: no remaining blocker. The export budget is now a useful CI gate rather than a reassuring report that can silently stop gating: every policy metric is required, finite, positive, and whole-numbered; malformed or empty policy files fail closed; an empty --json= value is rejected; and --json - keeps stdout machine-parseable by sending the human table to stderr. The route metric is named for the bootstrap resources it actually traverses rather than claiming to measure every cold-load byte.

All five prior review threads are addressed. Formatting, lint, type checking, 479 tests, production build, export verification, and budget measurement pass on this head.

mldangelo (commented):
Fresh exact-head re-review at `48de445aacfbc1ac41d0c75af220d0d75f440c66`.

**Disposition: no remaining blocker.** The PR is large for a personal site (1,421 added lines across the complete branch), so its maintenance cost deserves scrutiny. I think the capability is justified at this head: it measures the actual deploy artifact rather than only the JS graph; adds no dependency; shares the previously duplicated HTML/base-path logic with `verify-export`; runs quickly; and has 31 process-level fixture tests around the policy and CLI failure modes. The committed headroom keeps it a regression gate rather than a byte fingerprint.

This pass found and fixed four additional issues:

1. **A JSON report could invalidate its own measurement.** The script enumerated `out/`, then allowed `--json out/report.json`, creating a deployable file that was absent from the report. A second run would measure the stale first report and overwrite it with a differently sized report. JSON destinations inside `out/` are now rejected before measurement.
2. **The report could overwrite the CI policy.** `--json scripts/budget.json` read the valid limits and then replaced them with report data, guaranteeing a broken next run. The critical input path is now protected, with a test that verifies its bytes remain unchanged.
3. **Missing bootstrap assets made routes appear lighter.** A same-origin stylesheet/script/preload that did not resolve in the export was silently skipped. That could pair a broken page with a reassuringly smaller route metric. The gate now fails with the route and reference; a fixture removes the stylesheet and proves the failure.
4. **The SVG prose overstated exact savings.** `repeatedInlineSvgBytes` correctly counts full inline markup repeated past its first occurrence, but a sprite would not remove every one of those bytes: it still needs a definition and a `<use>` reference at each instance. The policy and PR description now call this an upper-bound deduplication opportunity, consistent with the [SVG 2 reuse model](https://www.w3.org/TR/SVG2/struct.html#UseElement).

The previously remediated policy behavior also remains sound: every metric is required, safe-integer, and positive on normal runs; unknown/malformed/empty policies fail closed; `--update-budget` alone can bootstrap; `--json=` is rejected; JSON stdout is machine-parseable via the documented silent npm form; and the route metric is explicitly the directly linked bootstrap set rather than all cold-load bytes. All five historical review threads remain resolved.

The PR metadata was materially stale and is now corrected: it no longer names `maxRouteFirstLoadBytes`, claims an absent metric remains ungated, reports 12/463 tests, describes a six-leg matrix that actually has four legs, or equates duplicate SVG markup with exact sprite savings.

Verification on the exact pushed head:

- focused measurer suite: 31/31 pass
- formatting, Biome (182 files), TypeScript, OG artifact check, and `git diff --check`: pass
- Vitest: 58 files / 482 tests pass
- coverage gate: pass (87.33% statements / 84.26% branches)
- exported-site integrity: 14 pages pass
- size gate: 186 files / 3,782,802 bytes; all seven limits pass at 82.7%–86.9% used
- hosted builds: Ubuntu Node 22/26, macOS Node 26, and Windows Node 26 all pass
- hosted quality, tests, aggregate CI, and Cloudflare Pages preview: pass

No browser-specific QA was necessary for the follow-up patch because it changes only post-build tooling, tests, policy documentation, and PR metadata. The fresh production build and export verifier confirm the site artifact still renders and links as before.

mldangelo:
Closing this rather than trimming it.

The gate contradicts its own stated invariant. `budget.json` and the AGENTS.md bullet both assert that no gated metric may be a function of page count — but `totalBytes` is gated, and by this PR's own measurements HTML (17.1%) plus RSC payloads (24.6%) means **~42% of the export scales linearly with routes**. The committed headroom (4,539,392 vs 3,782,802 measured) is about eleven more posts. Writing a twelfth post turns four build legs red for no defect, which is exactly the ratchet-without-reading failure the PR spends ~90 lines of prose promising to avoid.

`maxFileBytes` at 267 KiB has the same problem from the other direction: the first post embedding a screenshot larger than that fails CI for a correct and routine action.

Two of the seven budgets are also near-duplicates — `maxRouteBootstrapBytes` is approximately `documentBytes + cssBytes + fontBytes + javascriptBytes` given one shared stylesheet and one chunk graph, so it and `javascriptBytes` will always move together.

The underlying concern — nothing watches export size — is real but small. If it's worth revisiting, the shape that works is a handful of page-count-independent metrics (per-route JS, CSS, fonts) with no total-bytes gate, at roughly a tenth of the size. Not landing 1,531 lines to get there.

PR #936, opened 2026-07-28
fix(about): surface stated dates without inventing precision

mldangelo:
## Summary

- Turn the chronological About-page lists into a responsive log with year and age markers in a dedicated gutter.
- Read only dates and ages that the opening sentence actually states; do not derive a missing counterpart from `profile.birthDate`.
- Lift simple leading markers such as `At 7,` into the gutter, but keep inline and qualified wording such as `In the summer of 1996` intact.
- Leave undated entries undated instead of manufacturing precision, and keep the overlapping age range in chronological order.
- Avoid duplicate screen-reader announcements by hiding only gutter markers that repeat facts still present in the body. Lifted markers remain exposed.
- Preserve the layout at narrow viewports and in the printed page.

## Why

The existing lists contain useful chronology, but that information is difficult to scan. The log treatment makes explicit dates easier to find while treating the autobiographical prose as the source of truth. A year does not prove an exact age without the event date, and an age can span two calendar years, so the UI intentionally supports year-only, age-only, combined, and blank markers.

## Parsing behavior

- scans only the opening sentence
- recognizes anchored age/year phrases and age ranges
- ignores unrelated numbers such as `approximately 50 countries` and `Tandy 2000`
- preserves seasonal and other date qualifiers in the prose
- leaves entries that begin with rich content or contain no marker unchanged
- tests the real About content so new wording cannot silently corrupt the chronology

## Verification

- `npm run format:check`
- `npm run lint`
- `npm run type-check`
- `npm test` — 57 files, 461 tests
- `npm run og:check`
- `npm run build` — 16 static pages generated
- fresh exported HTML inspected for inline-versus-lifted accessibility behavior
- desktop, 320px mobile, and five-page print/PDF layouts inspected before the final accessibility-only delta


mldangelo:

mldangelo (commented):
Verdict: **changes needed before merge.** Filling genuinely missing markers is useful, and the parser handles the current prose without false-positive model numbers in the tested cases. But the new “one temporal system” manufactures precision the source text does not contain, drops meaningful wording, and still does not produce a chronological axis.

I ran the focused log-entry/Sections tests (26/26), the full suite (57 files, 462 tests), type-check, production export, and browser QA at 1100px. The layout is stable and does not overflow. The concerns are factual/editorial:

1. **Derived values are displayed as exact facts.** The profile birthday is February 5. For a year-only event such as “In 2016,” the person could be 25 before February 5 and 26 afterward, yet the gutter asserts `Age 26`. The travel entry may describe activity across more than one date, making the single exact age still less supportable. Conversely, an age-only event can fall in either of two calendar years depending on whether it happened before or after the birthday. The comment that the derivation is right “for all but the first five weeks” acknowledges uncertainty but the UI hides it. Tests currently pin that overclaim rather than validate a fact.

2. **The parser deletes `summer` from the MegaRace sentence.** “In the summer of 1996” becomes a bare `1996 / Age 6` marker plus prose with no season. That is a real loss of authorial meaning, not merely moving a date into the gutter.

3. **The displayed sequence is still not chronological.** The history currently moves through `2006`, then `2004–2007`, then `2008`. A consistent marker format is useful, but calling it one temporal “axis” implies order that the page does not provide. Either sort the content (after deciding whether changing narrative order is desirable) or describe this as consistent annotation rather than chronology.

I recommend rendering only the value explicitly supported by the prose. If a derived counterpart is truly valuable, mark it approximate (`~Age 26`) or represent the possible range. For age-only entries, avoid inventing a precise year without evidence. Preserve seasonal/qualifying text in the body or marker.

The two inline comments identify the lossy parse and the exact-value inference.

src/lib/logEntry.ts
```diff
@@ -2,57 +2,166 @@
...
+    pattern: /^I\s+was\s+(\d{1,2})\s+when\s+/i,
+    read: (m) => ({ age: [num(m[1])] }),
   },
-  // "In 2016, I visited Canada, Ethiopia, Austria…"
+  // "In 2016, I visited Canada …" / "In the summer of 1996, my uncle purchased …"
   {
-    pattern: /^In\s+(\d{4}),\s+/i,
-    label: (m) => m[1],
+    pattern: new RegExp(
+      `^In\\s+(?:the\\s+\\w+\\s+of\\s+)?(${YEAR_PATTERN.source}),\\s+`,
```
mldangelo: This pattern consumes `In the summer of 1996,` wholesale. The marker restores only `1996` (and a derived age), so the rendered entry loses “summer.” That changes the author’s factual wording. Capture/preserve the season in `rest`, or include it in the visible marker instead of treating the entire prefix as disposable.

src/lib/logEntry.ts
```diff
@@ -2,57 +2,166 @@
...
+/**
+ * Age and year are the same fact told two ways, so either can be derived from
+ * the other. The birthday falls in early February, so for all but the first
+ * five weeks of a year the age is simply the year minus the birth year; the
+ * gutter is an 11px annotation, and that is the precision it is asking for.
+ */
+function withDerivedValues({ age, year }: Reading): Reading {
+  return {
+    age: age ?? (year ? shift(year, -BIRTH_YEAR) : undefined),
+    year: year ?? (age ? shift(age, BIRTH_YEAR) : undefined),
```
mldangelo: These formulas turn partial information into an exact-looking fact. With a Feb. 5 birthday, a year-only event can occur at either of two ages, and an age-only event can occur in either of two calendar years. `year - birthYear` / `birthYear + age` is only an approximation without the event date. Please render only the stated value, visibly mark the derived value approximate, or model the possible range; a source comment acknowledging the error window does not make the UI claim accurate.

mldangelo:
Copilot's overview raised no actionable defect here.

One thing from this work is worth recording, because the owner review of #947 caught it: the source comment in `src/lib/logEntry.ts` claimed `Mavica MVC-FD71` is why `INLINE_YEAR` is anchored to a preposition. It cannot be — that string holds no four-digit run, so `YEAR_PATTERN` never matches it, and the example actually illustrates the *age* guard. AGENTS.md ended up carrying both the wrong claim and its own correction two lines apart.

Being fixed in #969: the comment now uses an example that genuinely reaches the year guard (`Tandy 2000`), each example verified against the real extractor, and the two AGENTS rules consolidated into one.

mldangelo:

AGENTS.md
```diff
@@ -100,6 +100,7 @@ Next.js 16 (App Router) · React 19 · TypeScript · Tailwind CSS v4 · Biome ·
...
+- **The about-page log has one temporal axis**: `src/lib/logEntry.ts` files every `Some History` / `Travel / Geography` entry under a year and prints the age beneath it, deriving whichever half the prose omits from `profile.birthDate` — the gutter used to print "Age 7" next to a sentence that said a year, which is two timelines in one column. It reads only the entry's opening sentence, and only the phrasings actually in the prose (`At 7,`, `14 - 17,`, `When I was 12,`, `I was 11 when`, `In the summer of 1996,`, and an embedded `…in 1995`). A leading phrase is lifted out of the sentence; an embedded one is annotated where it stands, because removing it would mean rewriting the author's words. So when a new entry dates itself some other way, add the form to `LEADING_PATTERNS` rather than rewording the entry — and keep the inline patterns anchored to a preposition or to "I was", or "approximately 50 countries" and "Mavica MVC-FD71" start becoming dates. `src/lib/__tests__/logEntry.test.ts` pins all fifteen history entries as dated and the log as opening on `profile.computingSince`
```
mldangelo: This durable rule currently institutionalizes three inaccuracies: the displayed list is not chronological (`2006 → 2004–2007 → 2008`), derived year/age counterparts are uncertain around the Feb. 5 birthday, and lifting `In the summer of 1996` deletes `summer`. The `Mavica MVC-FD71` example is also impossible for `YEAR_PATTERN` because it has no four-digit run. Rewrite this guidance only after the public semantics are corrected on this head.

mldangelo (commented):
Fresh exact-head re-review at `1b0be92022bc46854aa354c1a49bc1a3bcf587fa`.

**Verdict: changes needed before merge.** Filling genuinely blank history markers is useful and the desktop/mobile/print layouts are visually strong, but the “one temporal system” still converts partial autobiographical evidence into exact-looking facts, deletes source meaning, and does not actually produce a chronological axis.

The exact rendered output confirms the problems:

1. **Invented precision.** The profile birthday is February 5. A year-only entry such as “We subscribed to AOL in 1995” supports age 4 **or** 5, depending on the unknown event date, yet the gutter states `1995 / Age 5`. “In 2016, I visited…” can span many trips across a year, yet it states one exact `Age 26`. The inverse is equally uncertain: “At 7” spans portions of 1997 and 1998, but the UI states exactly `1997`. The comment at `withDerivedValues` admits a five-week error window; the public UI does not. Render only stated evidence, mark the counterpart approximate, or show the possible range.
2. **Loss of authorial meaning.** The source says “In the summer of 1996”; `LEADING_PATTERNS` consumes that whole phrase and the visible body now begins “My uncle purchased MegaRace…”. The gutter restores only `1996 / Age 6`, so “summer” is gone from screen, accessibility tree, and print.
3. **Not chronological.** The displayed “axis” runs `2006 → 2004–2007 → 2008`. Consistent annotation is not chronological ordering. Either sort after an explicit editorial decision about narrative order, or stop calling the device one temporal axis.
4. **Durable guidance is false.** `AGENTS.md:103` bakes in all three claims above and says `Mavica MVC-FD71` could become a year. It contains no four-digit run, so `YEAR_PATTERN` cannot match it; the author response correctly notes that this example belongs to the age guard. #969 may correct that comment on an integration branch, but #936 itself is unchanged and still carries the false rule.

The current parser tests mostly pin today’s output, including the overclaims. Add ambiguity-preserving cases around a non-January birthday and assert that qualifying words such as “summer” survive rendering; otherwise the suite protects the bug.

Independent exact-head QA:
- `format:check`, Biome (177 files), TypeScript, and `git diff --check`: pass
- Vitest: 57 files / 462 tests pass
- desktop semantic/browser inspection: all marker values and the nonchronological sequence confirmed
- 390×844 mobile inspection: stable, readable, no timeline overflow
- 5-page Letter PDF rendered and inspected; year/age pairs print cleanly inline, and the lost “summer” remains absent there too

The implementation is technically robust around extraction and layout, but autobiographical metadata must not be more precise than the author’s words.

mldangelo:

mldangelo (commented):
Final exact-head re-review at 7fe552b.

Disposition: no remaining blocker. The history log now preserves what the source actually says: “summer” is not discarded, year-only and age-only entries no longer acquire false exact counterparts, and entries are presented in defensible chronology. The maintenance guidance no longer institutionalizes the earlier inaccurate derivations or impossible pattern example.

I rechecked the timeline statements against the authored source text and the February birthday boundary. All three prior review threads are addressed. Formatting, lint, type checking, 461 tests, build/export, browser review, and printed-page QA pass.

mldangelo (commented):
## Exact-head re-review — `927aae68c955f47aa3f4578d219a36675e993dcd`

**Disposition:** no remaining blocking findings on this head.

I reviewed the full PR again for necessity, chronology and factual precision, parser behavior, accessibility, responsive/print presentation, public wording, tests, prior discussions, and PR metadata.

### Finding from this pass, now fixed

The inline-detection path intentionally keeps a stated fact in the sentence, but the gutter originally exposed the same fact again. A screen reader therefore encountered separate `1993` and `Age 3` labels immediately before prose that already said “in 1993 when I was 3”; the 1995 and qualified 1996 entries had the same duplication.

Commit `927aae6` now marks a gutter marker `aria-hidden` only when `extracted.rest === first`, which identifies the inline/copy case. A leading marker such as `At 7,` is removed from the body, so its gutter marker remains exposed. The real-content test pins both halves of that contract. This is an appropriate use of `aria-hidden` to remove redundant content from the accessibility tree while retaining the visible design; see the [WAI-ARIA guidance on hiding redundant or extraneous visible content](https://www.w3.org/TR/wai-aria-1.3/#aria-hidden).

### Design and factual assessment

- The change is useful: a dedicated gutter makes an already chronological autobiographical list materially easier to scan on screen and in print.
- The implementation now treats the prose as the source of truth. It displays only a year, age, or range the opening sentence actually states; it does not manufacture an exact counterpart from `profile.birthDate`.
- That restraint matters. A year-only event can span two ages around a February birthday, and an age can span two calendar years. The current UI honestly supports year-only, age-only, combined, and blank markers.
- Qualified wording is preserved: `In the summer of 1996` stays intact rather than being reduced to a bare year.
- The age-range entry precedes the age-16 entry, so the overlapping chronology is defensible.
- Pattern matching is narrow enough to avoid converting unrelated values such as `approximately 50 countries` or `Tandy 2000` into dates. Scanning only the opening sentence also prevents a later anecdotal year from relabeling the whole entry.
- The prose remains specific, personal, and natural. I found no grammar or phrasing issue worth flattening or rewriting merely for uniformity.

### Metadata and prior feedback

The former title/body described a rejected implementation: one derived year axis, inferred year/age counterparts, an unmodified `about.ts`, and 462 tests. I replaced them with an accurate explanation of the conservative source-first behavior and current 461-test result.

All three historical inline threads are resolved and outdated: the lost seasonal qualifier, invented date precision, and the durable guidance that encoded those errors are all corrected.

### Verification

Local, from a clean exact-head worktree except for the pre-existing untracked `node_modules` link:

- format check: pass
- lint: pass
- TypeScript: pass
- focused parser/component tests: 25/25
- full Vitest suite: 57 files, 461/461
- OG image consistency: pass
- production static build: pass, 16 routes generated
- export verifier: pass, 14 pages
- exported `/about/` HTML inspected: inline markers hidden from the accessibility tree; lifted markers exposed; `summer` retained
- desktop, 320px mobile, and five-page print/PDF presentation had already passed on the visual head; the final delta changes accessibility semantics only

Hosted on this exact SHA:

- Code Quality: pass
- Test: pass
- macOS / Node 26: pass
- Ubuntu / Node 22 and 26: pass
- Windows / Node 26: pass
- CI Success: pass
- Cloudflare Pages: pass

PR #937, opened 2026-07-28
feat(theme): render the control immediately and restore system mode

mldangelo:
## Summary

This fixes three user-facing gaps in the theme control and its surrounding navigation:

- the static header previously reserved a blank 44×44 hole until React hydrated
- the first click pinned light or dark in `localStorage`, with no route back to following the device and no subscription to later device-scheme changes
- when JavaScript was blocked, the page exposed enabled theme and mobile-menu buttons whose handlers could never run

The control now ships as a real button, supports `system → light → dark → system`, and follows `prefers-color-scheme` live while `system` is selected.

The parser-blocking bootstrap stamps `data-theme-choice` before the body paints. CSS uses that attribute as the progressive-enhancement boundary: if the bootstrap did not run, both client-only controls stay out of layout, tab order, and the accessibility tree. A 44px desktop action rail remains reserved so the primary navigation does not shift when scripts are unavailable. The footer remains the mobile no-JavaScript navigation.

## Design

The implementation keeps two distinct values on `<html>`:

- `data-theme`: the resolved appearance, `light` or `dark`
- `data-theme-choice`: the visitor's preference, `system`, `light`, or `dark`

That distinction is necessary because resolved dark cannot tell the control whether dark was forced or inherited from a dark device.

All three icon/label states are present in stable server markup. CSS exposes the state selected by `data-theme-choice`, so React does not conditionally replace the control during hydration. The accessible name identifies the preference and the next action, for example `Theme preference: system. Switch to light.`

The pre-paint initializer is a literal inline `<script>` in the document head. `next/script` with `beforeInteractive` was not sufficient here: App Router serialized the initializer into `self.__next_s`, leaving it for the client bootstrap to execute later. The exported HTML now contains the initializer itself, so persisted theme and choice are set during parsing rather than after client code loads. A layout test rejects a return to the queued descriptor form.

`src/lib/theme.ts` shares the attribute names, storage key, media query, choices, and transition behavior. The generated initializer must duplicate a small storage parser, so a test executes it across missing, valid, invalid, throwing-storage, and missing-`matchMedia` cases and compares it with the TypeScript reader instead of claiming the two paths cannot drift.

`system` is represented by the absence of the storage key. `usePrefersColorScheme` returns `null` until its first client read, subscribes to changes, and leaves the initializer's resolved theme untouched until the device preference is known.

The raw initializer is not intrinsically CSP-safe. This static export currently has no strict `script-src`; a future strict policy would need to allow the initializer's exact hash or permit inline scripts.

## Verification

- `npm run format:check`
- `npm run lint` — 182 files
- `npm run type-check`
- focused layout/theme/hook/control suite — 4 files, 43 tests
- `npm test` — 60 files, 489 tests
- `npm run og:check`
- `npm run build` — 16 routes generated
- exported HTML inspected: direct `<script id="theme-init">…</script>`, no `self.__next_s` wrapper, real button, and all three preference labels
- browser QA with JavaScript enabled at 1440×900 and 320×800: persisted dark survives reload, the full three-state cycle works, theme and menu targets remain 44×44, mobile menu opens, layout-shift entries are zero, and there are no hydration or console errors
- browser QA with scripts blocked: theme and hamburger controls have `display:none`, zero client rectangles, and no accessibility-tree or tab-order presence; desktop navigation remains exactly centered, and the footer loads `/about/` without JavaScript on mobile
- Cloudflare Pages preview for `efe2d59`: deployed successfully


mldangelo:

mldangelo (commented):
Verdict: **ready to merge; no blocking defect found.** This PR fixes two real product problems: the static HTML no longer reserves a blank theme-control hole, and a user can return to following the system preference instead of being permanently pinned after the first click.

I independently ran the targeted theme/hook/ThemeToggle tests (41/41), the complete suite (59 files, 486 tests), type-check, and the production static export. I also exercised the built control at desktop and mobile widths. There was exactly one visible theme state, the target remained 44×44, and there was no horizontal overflow. The actual cycle worked as described: dark → system (resolving to the current light OS preference) → light → dark, with the storage and accessible names tracking each transition. The exported HTML contains the real button rather than a placeholder.

The two-attribute model is justified: resolved appearance and user preference are distinct state, and trying to infer one from the other would make forced dark indistinguishable from system-dark. Keeping rendered markup stable across server and client is also the right approach for this static-export site.

One editorial refinement is worth considering, though I would not block on it. `Theme: system. Switch to light.` calls “system” the theme, when it is actually the preference and the resolved theme may currently be light or dark. Something like `Theme preference: system (currently light). Next: light.` is more semantically exact, though longer. The current label is still truthful enough and correctly communicates the action.

The implementation is large for a small control (878 additions across the PR), but much of that is defensive coverage of static markup, bootstrap failures, storage values, and media-query changes. Given that theme bootstrapping is flash- and hydration-sensitive, that coverage is valuable. I would only ask that the existing review nit about documenting the complete three-state cycle be cleaned up so the next maintainer does not have to reconstruct it from the transition helper.

mldangelo:
Implemented in #960, and then again properly in #967.

The docstring said `light → dark → system`; `THEME_CHOICES` in `src/lib/theme.ts` is `['system', 'light', 'dark']`, so the real cycle starts at `system`. Fixed.

The first fix then introduced a *fresh* false claim in the same sentence — it wrote the corrected order out by hand and asserted that order "is not a copy" and cannot drift, which is its own counterexample. #967 removes the hand-written sequence and points at `THEME_CHOICES` as the single source. Caught by the owner's review, not by me.

mldangelo:

src/components/Template/ThemeToggle.tsx
```diff
@@ -1,49 +1,129 @@
...
+ * "switch to dark mode" leaves a screen reader user unable to tell which of
+ * the three is active. The next state is derived from `nextThemeChoice` so the
+ * label cannot disagree with the cycle it describes.
+ */
+function labelFor(choice: ThemeChoice): string {
+  return `Theme: ${choice}. Switch to ${nextThemeChoice(choice)}.`;
+}
+
+/**
+ * Cycles light → dark → system, where `system` follows the device live.
```
mldangelo: This head still documents an incomplete/misleading cycle. `THEME_CHOICES` and `nextThemeChoice` define `system → light → dark → system`; the initial state matters because returning to system is the feature this PR adds. Point this docstring at the source of truth instead of hand-copying an order that has already drifted.

mldangelo (commented):
Fresh exact-head re-review at `47e968d7161d3b8c1031aa61954f0c7db6b164a8`.

**Verdict: ready to merge; no functional, accessibility, visual, or security blocker found.** This PR is necessary. It fixes two real user-facing gaps: the static header now contains a real 44×44 theme control instead of a hydration-time blank, and a visitor can return to following the system preference after choosing a forced theme.

The two-attribute model is justified. `data-theme` is resolved appearance; `data-theme-choice` is user intent. Trying to infer the latter from the former would make forced dark indistinguishable from system-dark. The pre-paint script and React path share constrained constants, storage values are accepted only when exactly `light` or `dark`, storage failures fall back safely, and the system media query is subscribed rather than sampled once.

I exercised the built site, not only the component tests:

- initial persisted light rendered `data-theme=light`, `data-theme-choice=light`, and accessible name “Theme: light. Switch to dark.”
- light → dark updated both attributes and the accessible name; reload preserved dark
- dark → system removed the forced choice, resolved to the current light device preference, and announced “Theme: system. Switch to light.”; reload preserved system
- system → light completed the cycle
- the same light → dark → system flow worked at 390×844; the icon remained visible beside the hamburger with no header or page overflow
- the exported HTML contains the real button and all three state labels; CSS exposes exactly one state

The only remaining defect is documentation: line 63 still says the component cycles `light → dark → system`, while the source of truth starts at `system` and cycles `system → light → dark → system`. The response says this was corrected in #960/#967, but #937’s head is unchanged. Please update the docstring here or merge through the integration head containing that correction. Prefer pointing to `THEME_CHOICES` rather than hand-copying the order again.

One nonblocking wording consideration: “Theme: system” names a preference as though it were a painted theme. “Theme preference: system (currently light). Next: light.” is semantically exact but longer. The current label remains understandable and correctly describes the next action.

Independent exact-head QA:
- `format:check`, Biome (181 files), TypeScript, and `git diff --check`: pass
- Vitest: 59 files / 486 tests pass
- desktop and mobile live interaction/persistence/accessibility QA: pass

The implementation is large for a small control, but the added coverage targets genuinely fragile prepaint, hydration, storage-failure, and media-query behavior rather than decorative complexity.

mldangelo:

mldangelo (commented):
Final exact-head re-review at f39c2f5.

Disposition: no remaining blocker. The theme control exposes system, light, and dark as real reachable states, retains the control in server-rendered HTML, and follows a live device-scheme change only while “system” is selected. Storage failures and pre-hydration behavior are covered. The misleading comment that enumerated a supposedly non-copied order was corrected so the documentation matches the shared state machine.

Targeted theme/bootstrap/toggle tests (19 in the final remediation pass), type checking, Biome, the full green CI matrix, and browser interaction checks support the result.

mldangelo (commented):
## Exact-head re-review — `1e4a2790e1ea921151a07b68f0d05a9586c8cc56`

**Disposition:** no remaining blocking finding on this head.

This pass did find a central defect that the previous exact-head review and component tests missed. It is fixed in `1e4a279`.

### First-paint blocker found and repaired

At the former head `f39c2f5`, `next/script` with `strategy="beforeInteractive"` did not emit the initializer as executable theme code. The production export contained:

```html
<script>
  (self.__next_s = self.__next_s || []).push([
    0,
    { children: "(function(){...})()", id: "theme-init" }
  ])
</script>
```

That only queued the IIFE for Next's client bootstrap. The exported `<html>` had neither theme attribute, so a returning forced-dark visitor could receive the default light first paint and system icon before the async client path ran. This directly contradicted the PR's core claim that theme and choice were settled before first paint.

The pre-hydration component test did not catch the transport failure because it manually stamped `data-theme-choice` before checking CSS. It proved that CSS could select the correct shipped label, not that the production document actually supplied the attribute in time.

The root layout now emits a literal parser-executed head script. A new render-level test requires the exact `themeInitScript()` body inside `<head><script id="theme-init">` and rejects `self.__next_s` wrapping. The fresh production export independently confirms the direct initializer.

This distinction is narrower than the general promise of `beforeInteractive`: the [Next.js Script guide](https://nextjs.org/docs/app/guides/scripts) guarantees execution before hydration, while this theme decision must occur during document parsing, before a first paint can use the default tokens.

### Other improvements from this pass

- Removed the false “CSP-safe” claim. This is an inline script in a static export with no nonce or hash. The [official Next.js CSP guide](https://nextjs.org/docs/app/guides/content-security-policy) explains that nonce-based CSP requires dynamic rendering and that static deployments need an allowed inline policy or hash-based approach. The durable repository guidance now states that constraint rather than implying protection that is not configured.
- Reworded the no-attribute CSS fallback. CSS cannot read storage, so the actual preference is unknowable if the initializer never runs; the old comment incorrectly claimed the page was following the device.
- Changed `Theme: system` to `Theme preference: system`. `system` is user intent, while the painted theme is resolved light or dark; the new accessible name accurately separates those concepts and still states the next action.
- Added a test that walking `nextThemeChoice` from `system` exactly equals `THEME_CHOICES`. The previous reachability test allowed the transition map and documented array order to disagree.
- Replaced the circular “A display, for the state that follows whatever the display is set to” comment with a direct description of the monitor icon.
- Replaced claims that generated and TypeScript paths “cannot drift.” They share vocabulary, but storage parsing is duplicated inside the generated string; the agreement test is what guards it.

### Design assessment

The PR remains necessary. Shipping a real 44×44 button removes a hydration-time hole, and restoring a reachable `system` state fixes a real persistence trap. The two-attribute model is justified: `data-theme` is resolved appearance; `data-theme-choice` is user intent. Forced dark and system-dark are otherwise indistinguishable.

The stable three-state markup is also the right design. CSS selects one state without React replacing the element, while React updates the root attributes and storage after hydration. The shared subscribed media-query hook fixes the former one-shot device read. Invalid/throwing storage and missing `matchMedia` fail safely.

The raw `dangerouslySetInnerHTML` use is constrained to a string generated from internal constants; no request, storage, or other user-controlled text is interpolated into the script source. It should nevertheless be revisited if a strict CSP is introduced, as the updated guidance says.

### Verification

Local exact-head evidence:

- format check: pass
- Biome: 182 files pass
- TypeScript: pass
- focused layout/theme/hook/control suite: 4 files, 43/43
- full Vitest suite: 60 files, 488/488
- OG consistency: pass
- export verifier: pass, 14 pages
- emitted HTML: direct `theme-init` IIFE, no queued descriptor, real button, all three preference labels
- browser QA: persisted dark survives reload; dark → system exposes the system state and follows the dark device; exactly one icon is visible; target is 44×44; no document overflow

Hosted on this exact SHA:

- Code Quality: pass
- Test: pass
- macOS / Node 26: pass
- Ubuntu / Node 22 and 26: pass
- Windows / Node 26: pass
- CI Success: pass
- Cloudflare Pages: pass

Both historical cycle-documentation threads are resolved and outdated. The PR title and body have been rewritten to describe the final implementation and current 488-test result rather than the superseded queued-script behavior.

mldangelo (commented):
Follow-up exact-head re-review of `efe2d59b5c5d375f9182fd7a6f6c747d78314c7a` against `main` at `e638bce514f8965d42e0279fb23c4a007df529e2`.

**Disposition: no remaining blocker.** The feature is necessary and the current implementation is materially better than the blank pre-hydration slot and permanent two-state preference it replaces. This pass did, however, find and fix a progressive-enhancement defect the previous review missed.

With JavaScript blocked, the initializer never stamped `data-theme` or `data-theme-choice`, so the page correctly remained light. The UI nevertheless exposed two enabled controls that could not work:

1. the theme button announced “Theme preference: system. Switch to light” even though the page was already light and no handler could run;
2. at mobile widths, “Open navigation menu” remained in the accessibility tree and tab order even though the hidden menu could not open.

The fix makes the pre-paint attribute the CSS enhancement boundary:

- `.theme-toggle` and `.hamburger-button` are hidden by default;
- `:root[data-theme-choice]` reveals them after the parser-blocking initializer proves scripts ran;
- the desktop `.nav-actions` rail keeps a 44px minimum width, so removing the no-JS control does not shift the centered primary navigation;
- focused tests exercise both raw DOM presence and accessibility-role absence before the attribute, then visibility and role discovery after it;
- the repository guidance now records that client-only controls must not promise actions when their handlers cannot hydrate.

Fresh verification on this exact head:

- `npm run format:check`: clean
- `npm run lint`: 182 files, clean
- focused navigation/theme suite: 35 tests passed
- `npm test`: 60 files, 489 tests passed
- `npm run og:check`: current
- `npm run build`: passed, 16 static routes
- `npm run verify-export`: 14 pages passed
- Cloudflare Pages preview: deployed successfully for `efe2d59`
- JavaScript enabled at 1440×900 and 320×800: correct pre-hydration name, full system → light → dark → system cycle, persisted dark reload, 44×44 targets, working mobile menu, zero layout-shift entries, no overflow, and no console or hydration errors
- scripts blocked: both client-only controls have `display:none`, zero client rectangles, and no accessibility-tree or sequential-tab presence; desktop primary navigation remains centered at exactly 720px; the mobile footer navigation loads `/about/` without JavaScript

One deliberate tradeoff remains: if the tiny inline initializer runs but the React bundle subsequently fails, the attribute can reveal controls whose handlers never hydrate. Avoiding that rare partial-failure state would require giving up the PR's core pre-hydration-control goal or duplicating behavior outside React. It is not a blocker for the normal or true no-JavaScript paths verified here.

Both existing inline review threads are resolved and outdated.

PR #938, opened 2026-07-28
fix(stats): measure site readings and disclose their sources

mldangelo:
## Why

The stats page mixed three different kinds of claim—repository measurements, GitHub API values, and maintained profile facts—without saying which was which. It also published a hand-entered `Number of linter warnings: 0`, even though the page cannot observe a CI run. That made a page about measurement both hard to read and factually brittle.

## What changed

- Count repository figures at build time instead of typing them into the data file:
  - direct dependencies from `package.json`;
  - installed non-development package locations from npm’s `node_modules/.package-lock.json` for the current build host;
  - package locations represented in `package-lock.json`;
  - explicitly enabled Biome rules from `biome.json` and its overrides.
- Return `null` when a manifest is absent or invalid, and omit that row rather than inventing a value.
- Resolve and format both stats tables through `src/lib/readings.ts`, with pinned `en-US` number formatting and UTC formatting for the GitHub push date.
- Label each substantive row as `Measured`, `GitHub`, or `Profile`; visibly distinguish a stale GitHub fallback build.
- Keep package-location wording precise: these counters measure lockfile/install locations, not unique package names.
- Remove internal file paths from public copy and avoid linking the host-specific installed-tree count to the portable lockfile.
- Preserve the existing compact table design and row-header semantics; provenance sits below each label so values keep their first-baseline alignment.

The joke row, `Number of spoons`, remains intentionally unmarked because it measures nothing.

## Verification

- `npm run format:check`
- `npm run lint`
- `npm run type-check`
- `npm test` — 499 tests in 59 files
- `npm run og:check`
- `npm run build` — 16 static routes generated; the local sandbox correctly exercised the dated fallback because outbound DNS was unavailable
- 320px browser QA — no horizontal overflow or console errors; revised labels and fallback/provenance notes rendered correctly

The installed non-development counter also matched `npm ls --omit=dev --all --parseable` after excluding the project root on the audited build host.


mldangelo:

src/lib/manifest.ts
```diff
@@ -0,0 +1,176 @@
...
+    total += 1;
+
+    if (!isRecord(entry)) {
+      continue;
+    }
+
+    if (
+      entry.dev === true ||
+      entry.devOptional === true ||
+      entry.optional === true
```
mldangelo: **Blocking — this drops packages that npm installs in production.** The lockfile’s `optional` flag means the package is strictly in an `optionalDependencies` tree; it does not mean “wrong-platform binary.” npm’s production default omits `dev`, while omitting `optional` requires a separate choice. On this exact checkout the function returns **54**, but `npm ls --omit=dev --all --parseable` with Node 26.5/npm 11.17 returns 71 paths including the root (**70 installed dependencies**), including compatible optional packages such as `@next/swc-darwin-arm64` and `sharp`. Please do not publish 54 as the production tree. Either measure an actual platform install or remove/reframe this metric. See npm’s [lockfile field semantics](https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json/#packages) and [`omit` defaults](https://docs.npmjs.com/cli/v11/commands/npm-install/#omit).

src/data/stats/site.ts
```diff
@@ -1,55 +1,106 @@
...
+    // actually wants: what the project chose, what that resolves to, and how
+    // much of it a production install carries.
+    label: 'Dependencies declared directly',
+    key: 'direct_dependencies',
+    source: 'measured',
+    unit: 'packages',
+    link: `${BLOB}/package.json`,
+  },
+  {
+    label: 'Installed to run in production',
```
mldangelo: **Blocking copy/meaning issue:** this is an `output: 'export'` site. Production serves static files from `out/`; it does not install packages to “run” the site. Even a corrected npm-tree count would describe a build environment, not the deployed product. Please remove this row or rename/redefine it to an honest, useful artifact metric (for example, exported asset count/weight). As written, the confident label makes the undercount above more misleading.

mldangelo (commented):
## Verdict: changes required

The provenance work, fallback styling, thousands separators, and removal of the hand-typed “0 linter warnings” claim are useful improvements. The new manifest subsystem is not ready to publish, however: its most prominent new number is both semantically misleading for this static-export site and quantitatively wrong on the platform I tested.

### Blocking findings

1. **“Installed to run in production” is not what the counter measures.** This repository publishes a static `out/` directory; no Node dependency tree is installed to *run* the production site. At best this could describe packages installed to *build* the artifact on one platform, but the implementation does not measure that either.
2. **The optional-package filter undercounts the actual production dependency tree.** npm documents `optional` as “strictly part of the optionalDependencies tree,” not “a binary for some other platform,” and `optional` is only omitted when explicitly requested. On this exact head, `countLockedPackages().production` returns **54**; with Node 26.5/npm 11.17 the installed tree reports **71 paths including the root = 70 dependencies** via `npm ls --omit=dev --all --parseable`. Compatible optional packages such as `@next/swc-darwin-arm64`, `sharp`, and its Darwin binaries are present. Official npm references: https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json/#packages and https://docs.npmjs.com/cli/v11/commands/npm-install/#omit.

I recommend removing this row unless it can be replaced by a metric that describes the deployed static artifact (for example, exported asset count/weight). If package-tree complexity is genuinely useful to visitors/forkers, label the build context precisely and derive the value from an actual clean install for the build platform rather than inferring it by deleting every optional lock entry.

### Scope / design assessment

This PR adds 1,068 lines across 18 files, substantially to support four package/lint rows on a personal “fun numbers” page. I think the smaller core—shared resolution, provenance, honest fallback presentation, and count formatting—does improve the site. The dependency/lockfile subsystem needs a stronger user-facing purpose to justify its maintenance and visual weight. Four adjacent build-internals rows currently read more like repository diagnostics than portfolio content.

Copy nits worth addressing while revising: use the GitHub brand capitalization in “Stars this repository has on GitHub” and “Open GitHub issues and pull requests”; “Last updated at” is less precise than “Last pushed” for `pushed_at`.

### Independent verification at `8837579c685759c3eeb132cd5f372b1b93e94c79`

- Targeted stats/manifest/readings tests: **80 passed**
- Full suite: **494 passed / 59 files**
- `npm run format:check`: pass
- `npm run lint`: pass (181 files)
- `npm run type-check`: pass
- `npm run build` on Node 26.5: pass; 16-page static export, `/stats` prerendered
- The build exercised the documented GitHub fallback because sandbox DNS was unavailable; generated HTML correctly disclosed the fallback
- Browser QA of the built export at 320px: no horizontal page overflow; all provenance marks remained single-line and the table rows stayed within the viewport
- Existing Copilot feedback already identifies the absolute “component is never re-rendered” comment; I have not duplicated that inline

The tests pass because their fixtures encode the same blanket optional exclusion; none compares the claimed number with npm’s installed production tree.

mldangelo:
Copilot's inline note on `Personal.tsx` was correct in principle — a component can re-render when its parent does, so "never re-rendered" is not a claim React supports.

It turned out to be moot: that comment no longer exists on the integration branch. It was deleted when `Personal` became a server component again as part of rendering the age with the power off, and replaced with a docstring about that. Confirmed by reading the file and by `git log -S`.

The related claim that *did* survive is fixed in #968: `Site.tsx` still said "no client-side JavaScript shipped" while this very branch added a `BuildClock` client leaf. The accurate property is narrower and more interesting — the client JS on `/stats` is confined to two leaves that write out of band, so the table itself stays static.

Also in #968: `github` → `GitHub` in every user-visible label, without touching the API response keys.

mldangelo:

mldangelo (commented):
## Fresh exact-head re-review — changes still required

I re-reviewed the complete current diff, the earlier Copilot and owner feedback, the author's follow-up, the generated export, and the rendered page. The current PR head is **unchanged** from the prior independent review, and no commit on this PR addresses its two substantive blockers.

### Current-head findings

1. **The production-package count is still quantitatively wrong.** On this exact checkout, the generated page publishes **54 packages**. A fresh npm ls --omit=dev --all --parseable under the repository's Node 26.5 baseline reports **71 paths including the root, i.e. 70 installed dependencies**. The implementation still drops every lockfile entry marked optional; npm does not omit optional dependencies from a normal production install unless optional is separately omitted. Compatible packages such as the platform SWC and Sharp binaries are therefore excluded from the displayed number. The passing fixtures only reproduce the implementation's assumption; they do not compare it with npm's installed tree.

2. **“Installed to run in production” is still the wrong concept for this site.** This project deploys the static out/ artifact. No Node dependency tree is installed to run the production site, so even a corrected platform-specific npm count would describe a build/install environment rather than the deployed product. This remains a prominent, confident, user-facing misstatement. Removing the row, or replacing it with an exported-artifact measurement such as file count/weight, would be more useful and honest.

The author's July 30 response notes that comment/copy cleanup exists on the integration branch or in #968, but those changes are not in this PR's current head. On the exact code under review, PersonalStats still claims it is “never re-rendered,” and SiteStats still says “no client-side JavaScript shipped.” Those are non-blocking code-comment inaccuracies here, but they reinforce that integration-branch state must not be used as evidence that this PR itself is ready.

### Necessity and design

The useful core remains worthwhile: removing the hand-typed warning claim, sharing reading resolution, showing provenance, formatting large values, and visibly flagging fallback GitHub data all improve /stats. The mobile presentation is clear and stable. The package/lockfile subsystem, however, accounts for a large portion of a 1,068-line / 18-file change while adding repository diagnostics that are less meaningful to a portfolio visitor than the existing personal/site readings. I would keep the provenance/formatting work and either remove or sharply redefine the dependency metrics.

Previously noted copy remains unchanged: user-facing “github” should be **GitHub**, and pushed_at is more accurately labelled **Last pushed** than “Last updated at.”

### Fresh verification at 8837579c685759c3eeb132cd5f372b1b93e94c79

- Exact diff check: clean
- Targeted stats/manifest/readings suite: **80 passed / 7 files**
- Full suite: **494 passed / 59 files**
- npm run format:check, npm run lint, and npm run type-check: pass
- npm run build: pass; 16 static routes
- npm run verify-export: pass; 14 pages OK
- Generated /stats HTML independently inspected: publishes “Installed to run in production — 54 packages”
- Fresh npm-tree comparison: **70 installed dependencies** excluding the root
- Browser QA at 320 px: no horizontal overflow; both tables stay within 24–296 px; provenance marks remain legible and the fallback state is visibly disclosed

Because the head did not change, I have not duplicated the existing inline threads. This review confirms those blockers against a fresh clean build rather than merely carrying the previous verdict forward.

mldangelo:

mldangelo (commented):
Final exact-head re-review at 11241df.

Disposition: no remaining blocker. The dependency reading now measures the installed non-development tree from node_modules rather than misinterpreting package-lock optional flags, and the public label describes a build-host dependency count rather than pretending a static export installs packages at runtime. The React comment now promises only the meaningful invariant: the ticking age itself does not trigger renders.

All three prior review threads are addressed. Formatting, lint, type checking, 498 tests, build/export checks, and the installed-tree comparison pass on this head.

mldangelo (commented):
Re-reviewed the full diff at exact head `3f76249d6253479f6deae21365e840c797a0c383`, including the prior review threads, the rendered export, the installed-tree semantics, and the public copy.

## Disposition

No unresolved blocker remains at this head. This PR is necessary: the page previously presented a hand-entered `0` as an observed lint result and rendered measured, API-sourced, and maintained profile facts as if they had identical provenance. The replacement makes the claims inspectable and improves the site rather than adding a decorative metric layer.

## Findings from this pass and how they were resolved

- **Host-dependent date bug:** `dayjs(...).format(...)` used the build host’s local time zone. For example, `2024-01-15T00:30:00Z` rendered as January 14 on a Pacific-time host. The push date is now formatted with an explicitly pinned `en-US` locale and `UTC` time zone, with a boundary regression test.
- **False CI wording:** “Enforced by CI on every push” overstated the workflow; branch pushes other than `main` do not trigger it. The row now says exactly what is measured: `Biome lint rules enabled in CI`.
- **Wrong evidence link:** the installed-tree count comes from the host-specific `node_modules/.package-lock.json`, but its value linked to the committed, cross-platform `package-lock.json`. That link is removed; the portable lockfile count retains its own link.
- **Package-count imprecision:** both installed-tree and lockfile counters count package locations, not unique package names (63 locations versus 62 names locally; 301 versus 297). The UI now says `Installed non-development package locations` and `Lockfile package locations` rather than implying unique packages.
- **Incomplete provenance claim:** the source note said all measured rows came from the working tree even though one comes from the installed dependency tree. Both live and fallback notes now describe the checkout *and* installed tree.
- **Incorrect internal invariant:** the `StatDeclaration` documentation said no keyed declaration may carry a literal, while keyed profile rows intentionally do. The prohibition is now correctly scoped to measured site rows; supplied measurements still win when a caller deliberately overrides a maintained profile value.
- **Public implementation leakage and absolute wording:** the personal note no longer exposes `src/data/profile.json`, and the server-component comment now claims only that this table adds no client JavaScript—not that the whole page ships none.
- **Mobile copy pressure:** the precise package-location labels were shortened without losing meaning. At 320px, the table remains within a 320px document width and the revised rows render without horizontal overflow.

The npm model behind the counters is now aligned with the official documentation: the hidden lockfile represents the installed `node_modules` tree when current, its `packages` object maps package *locations*, and `devOptional` can describe an optional dependency of a non-development dependency rather than a purely development-only node. See [npm’s hidden-lockfile and package-map documentation](https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json/#hidden-lockfiles) and [npm’s `omit` documentation](https://docs.npmjs.com/cli/v11/commands/npm-install/#omit). The local counter also matched `npm ls --omit=dev --all --parseable` after excluding the project root.

## Verification

- Format, Biome lint, TypeScript, and OG artifact checks passed.
- Full Vitest suite: **499/499 tests across 59 files**.
- Production build generated all 16 static routes; `verify-export` passed all 14 exported pages.
- The sandboxed local build deliberately exercised the dated GitHub fallback, which rendered with the corrected UTC date and explicit fallback note.
- Browser QA at 320px found no document overflow or console errors; provenance marks, source notes, and unlinked host-specific count were present in the generated HTML.
- Hosted exact-head checks are green: Code Quality, Test, Node 22/26 Ubuntu builds, Node 26 macOS/Windows builds, CI Success, and Cloudflare Pages.
- Independent security and accessibility passes found no issue at this head.

The PR title and body were also rewritten to remove the obsolete rejected implementation (`54 production packages`, `package-lock.json` as the installed-tree source, and 494-test evidence) and now describe this exact head.

PR #939, opened 2026-07-28
feat(resume): publish a machine-readable JSON Resume

mldangelo:
## Why

The résumé data claimed JSON Resume compatibility, but the site did not publish a JSON Resume document. This change adds a useful, portable representation for résumé tools without introducing a second hand-maintained résumé.

## What changed

- Adds a force-static `/resume.json` route assembled from the same profile, work, education, course, contact, and skill data used by the site.
- Adds a visible `JSON` link on `/resume/` and an HTML `<link rel="alternate" type="application/json">` discovery hint.
- Converts the rich work summaries to plain text for the JSON consumer while preserving the source markup used by the HTML résumé. The conversion strips supported anchors and Markdown formatting without treating comparisons such as `latency < 50ms` or generic types such as `Map<string, number>` as HTML.
- Extends `verify-export` so the build fails closed when the artifact is missing or malformed, its root is not an object, required name/work values have the wrong shape, unsupported root keys appear, summaries contain markup, site URLs are invalid, or the résumé page stops linking and advertising the JSON document.
- Documents the mapping and export invariants in `AGENTS.md` and adds focused route, mapping, metadata, and fault-injection tests.

## Mapping decisions

- `basics` comes from the shared profile, contact, and site metadata.
- Work entries come from the rendered résumé data; inline links are flattened to readable text only in the JSON output.
- Degrees and courses are grouped under the corresponding education entries.
- Skills remain category-based. Numeric competency values determine keyword order rather than being translated into unsupported qualitative claims.
- Empty schema sections and facts the source data does not establish—such as a country code, proficiency labels, or a synthetic `lastModified` value—are omitted.

The root keys and field shapes were checked against the pinned [JSON Resume v1.0.0 schema](https://raw.githubusercontent.com/jsonresume/resume-schema/v1.0.0/schema.json) and the [JSON Resume schema documentation](https://jsonresume.org/schema/). The schema rejects unknown root properties, but it also validates nested types, URI/date formats, and other field constraints; the repository’s export checks supplement rather than replace that validation.

## Validation

- `npm run format:check`
- `npm run lint` — 182 files
- `npm run type-check`
- `npm run og:check`
- `npm test` — 60 files, 489 tests
- `npm run build` — 17 static routes, including `/resume.json`
- `npm run verify-export` — 14 pages plus feed, sitemap, and résumé checks
- Independent Draft 4 validation of `out/resume.json` against the official pinned v1.0.0 schema — valid
- Static HTTP check — `Content-Type: application/json`, 10,599-byte response
- Browser QA at 320px — visible aligned affordance, no horizontal overflow, console errors, or page errors

No server-only feature or runtime dependency is added; the JSON document is emitted by the existing static export.


mldangelo:

app/resume/page.tsx
```diff
@@ -26,7 +27,20 @@ export default function ResumePage() {
...
       <section className="resume-page">
         <header className="resume-header">
-          <h1 className="resume-title">Resume</h1>
+          <div className="resume-header-row">
+            <h1 className="resume-title">Resume</h1>
+            {/* The same affordance as the RSS chip on /writing: this résumé is
+                also published as data, at the JSON Resume path the work data
+                has always claimed to conform to. */}
+            <a
+              href={RESUME_JSON_PATH}
```
mldangelo: **Blocking for forks:** this raw root-relative anchor is not rewritten when `actions/configure-pages` injects a repository `basePath`. On a fork hosted at `/personal-site/`, it points to the domain-root `/resume.json` instead of `/personal-site/resume.json`. The new verifier’s base-path fixture manually supplies the expected prefixed HTML, so it does not exercise this component and gives false confidence. Use a base-path-aware link strategy (Next documents that `next/link` prefixes `basePath`, or use an intentionally relative file URL) and add an assertion against a real base-path build. See [Next’s basePath link behavior](https://nextjs.org/docs/pages/api-reference/config/next-config-js/basePath#links).

src/lib/resumeJson.ts
```diff
@@ -0,0 +1,290 @@
...
+ * a test.
+ */
+function lastModified(): string {
+  const dates = work.flatMap((position) =>
+    [position.startDate, position.endDate].filter(
+      (date): date is string => !!date,
+    ),
+  );
+  // ISO dates sort lexicographically.
+  return `${dates.sort().at(-1)}T00:00:00Z`;
```
mldangelo: **Blocking semantics:** this is the latest employment event, not when the JSON Resume was last modified. Editing `profile.json`, a skill, education, any summary/highlight, or this generator leaves the published value unchanged; this brand-new artifact therefore claims `2026-03-09T00:00:00Z`. Because JSON Resume’s `meta.lastModified` is optional, please omit it unless it can be derived from the actual document/source revision. A deterministic but knowingly unrelated timestamp is worse than no timestamp.

mldangelo (commented):
## Verdict: changes required

This is a worthwhile feature: a first-class JSON Resume artifact is genuinely useful, it is sourced from the same data as the page, and the generated document validates against the pinned official JSON Resume v1.0.0 schema. The current head still has three release blockers.

### Blocking findings

1. **The new affordance breaks on fork/repository-site deployments.** The page emits a raw `href="/resume.json"`. This repository deliberately relies on `actions/configure-pages` injecting a Next `basePath` for forks. Next’s documentation says the base path is automatically applied to `next/link`; a native anchor is left untouched. A fork served at `/personal-site/` will therefore navigate to the domain root rather than `/personal-site/resume.json`. The base-path verifier fixture cannot catch this because it manually writes the already-correct `href="${basePath}/resume.json"` rather than rendering `ResumePage`. Official Next reference: https://nextjs.org/docs/pages/api-reference/config/next-config-js/basePath#links.
2. **Malformed numeric entities can crash the build.** I confirmed the existing Copilot finding on this exact head: `toPlainText("bad &#999999999;")` reaches `String.fromCodePoint(999999999)` and throws `RangeError: Invalid code point 999999999`. The range and surrogate interval need validation, with the original entity preserved when invalid. Please add regression cases; I did not duplicate Copilot’s inline thread.
3. **`meta.lastModified` is not the document’s last-modified time.** It is the maximum employment start/end date. A change to the profile, education, skills, summary wording, or generator leaves it at `2026-03-09T00:00:00Z`; indeed this newly generated artifact claims it was last modified months before it existed. JSON Resume makes the field optional, so omission is more accurate unless it can be derived from the actual content/source revision.

### Scope / design assessment

The feature improves the site and is justified. The JSON chip is readable and unobtrusive at 320px, does not create horizontal overflow, has a clear accessible name, and is correctly suppressed in print CSS. The generated JSON is well structured and readable. The implementation is heavier than it needs to be in one place: `verify-export.mjs` and its tests duplicate the schema’s root-key list while only partially validating the document. Since the document advertises `$schema`, a real validation against the pinned schema would provide stronger evidence; otherwise describe this gate as artifact-integrity checks rather than schema validation.

### Independent verification at `eff9d7201e78844176a570384e65db7879f4b58e`

- Targeted route/affordance/resume/export tests: **47 passed / 4 files**
- Full suite: **481 passed / 60 files**
- `npm run format:check`: pass
- `npm run lint`: pass (182 files)
- `npm run type-check`: pass
- `npm run build` on Node 26.5: pass; 17 routes generated, including static `/resume.json`
- `npm run verify-export`: pass (14 HTML pages plus resume artifact checks)
- Independently validated `out/resume.json` with Python jsonschema Draft 4 against the downloaded official v1.0.0 schema: **pass**
- Confirmed generated HTML contains `href="/resume.json"`, which is correct at the canonical root deployment but demonstrates the unprefixed fork path
- Browser QA at 320px: JSON control remained within the title row (right edge 296px in a 320px viewport), no horizontal overflow, accessible name “JSON Resume”
- Print rule audited: `.resume-json-link` is hidden alongside the other navigation affordances

mldangelo:
Implemented in #963. This was the most serious item in the batch and it reproduces exactly as described:

```
$ node -e "String.fromCodePoint(parseInt(\"99999999\",10))"
RangeError: Invalid code point 99999999
```

The old guard only checked `Number.isNaN`, which does not catch an out-of-range value, so `&#99999999;` in any `summary` string in `src/data/resume/work.ts` would have failed the production build with an unhandled `RangeError`. The code point range is now validated before `String.fromCodePoint`, falling back to leaving the entity text intact — the same behaviour the function already had for an unknown named entity.

Lone surrogates (`0xD800`–`0xDFFF`) are handled deliberately too: `String.fromCodePoint` does *not* throw for those, it emits an unpaired code unit that then travels into JSON, so those are left intact rather than emitted.

mldangelo:

mldangelo (commented):
## Fresh exact-head re-review — changes still required

I re-read the full current diff, the prior Copilot/owner review threads, the author's July 30 response, the official JSON Resume v1.0.0 schema, the generated artifact, the responsive page, and a newly rendered five-page print PDF. The feature is still useful and the emitted JSON is schema-valid, but this PR's head is unchanged and all three release blockers remain on the exact branch being reviewed.

### Current-head blockers

1. **The new link still breaks repository-site forks.** The real exported Resume page contains href="/resume.json". A document served from https://owner.github.io/personal-site/resume/ therefore navigates to https://owner.github.io/resume.json, not /personal-site/resume.json. Next's current [basePath documentation](https://nextjs.org/docs/pages/api-reference/config/next-config-js/basePath) explicitly says next/link applies the configured prefix; this native anchor does not. The verifier's base-path fixture remains false confidence because it manually writes an already-prefixed anchor instead of rendering ResumePage.

2. **Malformed numeric entities can still crash the production build.** The unchanged decoder still passes any non-NaN integer to String.fromCodePoint. Freshly exercising String.fromCodePoint(parseInt("999999999", 10)) under the repository's Node baseline throws RangeError: Invalid code point 999999999. The author's response correctly acknowledges this and says the guard was implemented in #963, but that fix is not present in this PR's head. Out-of-range values and the surrogate interval need to be rejected here, with the original entity preserved.

3. **meta.lastModified still describes an employment event, not document modification.** The newly generated out/resume.json still claims 2026-03-09T00:00:00Z because that is the newest work date. Editing profile data, education, skills, any prose, or the generator leaves that value unchanged. The [official v1.0.0 schema](https://raw.githubusercontent.com/jsonresume/resume-schema/v1.0.0/schema.json) makes meta and lastModified optional; omitting the field is more accurate than publishing a deterministic but unrelated timestamp.

Fixes that exist only on an integration branch or follow-up PR do not make this exact PR safe to merge independently. If these branches are intended to land solely through an integration PR, that dependency should be explicit and GitHub should not present this branch as independently ready.

### Necessity and design

A first-class JSON Resume artifact is justified: it creates a useful machine-readable output from data the site already maintains, and the small JSON affordance is visually restrained, readable, keyboard-accessible, and correctly excluded from print. The assembly code preserves the page's inline links while producing plain text, and the official schema validation passed.

The export verifier is useful as an artifact-integrity gate, but it should not be described as schema validation: it duplicates selected root keys and checks a subset of invariants. Either pin and run the actual schema in CI, or name this layer accurately.

### Fresh verification at eff9d7201e78844176a570384e65db7879f4b58e

- Exact diff check: clean
- Targeted route/affordance/resume/export tests: **47 passed / 4 files**
- Full suite: **481 passed / 60 files**
- npm run format:check, npm run lint, and npm run type-check: pass
- npm run build: pass; 17 static routes including /resume.json
- npm run verify-export: pass; 14 pages plus résumé artifact checks
- Fresh Draft 4 validation against the official JSON Resume v1.0.0 schema: **valid**
- Generated artifact: 10 work entries; canonical URL present; meta.lastModified remains 2026-03-09T00:00:00Z
- Generated HTML: href="/resume.json" confirmed
- Browser QA at 320 px: no overflow; JSON control remains within 236–296 px and has the accessible name “JSON Resume”
- Print QA: rendered all five letter-size pages to PNG and inspected each; the JSON chip is correctly absent. The existing skip-link print defect is visible in that baseline output, but it is not introduced by this PR.

Because the code did not change, I have not duplicated the existing inline threads. This is a fresh current-head confirmation with new build, schema, browser, and PDF evidence.

mldangelo:

mldangelo (commented):
Final exact-head re-review at 913da3c.

Disposition: no remaining blocker. /resume.json is now a useful machine-readable companion to the page: its link is repository-base-path safe, numeric HTML entities are decoded only when they are valid Unicode scalar values, the document validates against the intended schema boundary, and the misleading employment-derived meta.lastModified claim is omitted rather than published as document provenance. Serialization is also tested off the wall clock.

All three prior review threads are addressed. Formatting, lint, type checking, 483 tests, production build, export verification, and the root/repository-path cases pass.

mldangelo (commented):
## Re-review of exact head `fb22e80672583b25df908128d984536b6673a489`

**Disposition: no remaining blockers on this head.** This is a justified site improvement: the repository already described its résumé data as JSON Resume-compatible, and a static, discoverable `/resume.json` gives people and résumé tooling a portable representation without adding a parallel source of truth or a runtime service.

### Problems found in the prior head and fixed here

1. **The export verifier accepted structurally invalid JSON documents.** A falsey primitive such as `null` could pass the old parse guard, while truthy but invalid values such as `basics.name: []` and `work: [1]` were not rejected. The verifier now requires an object root, a non-empty string name, and object work entries. Fault-injection tests cover each case.
2. **The plain-text converter treated every angle-bracket span as HTML.** The broad `<[^>]*>` expression could silently delete legitimate résumé text such as `latency < 50ms` or `Map<string, number>`. It now unwraps only the supported anchor tags; a regression test pins comparisons and generic types, and unsupported markup remains visible to the export verifier instead of disappearing.
3. **The JSON document was visible but not machine-discoverable.** `/resume/` now emits `<link rel="alternate" type="application/json" href="https://mldangelo.com/resume.json">`. Metadata and export tests verify both the canonical URL and the alternate, including repository-site base paths.
4. **One route test overstated what it proved.** The route unit test verifies the response header produced before export; it does not prove which header GitHub Pages ultimately serves. The test name now states that boundary, while a separate static HTTP check verifies the emitted `.json` file is served as `application/json`.
5. **Documentation and PR claims were too absolute or stale.** Shared source data reduces duplicate maintenance but does not make drift impossible; JSON Resume validates more than unknown root keys; and the previous test/byte counts no longer matched the head. `AGENTS.md`, the title, and the PR description now describe those limits and the current evidence precisely.

### Design and content review

- The force-static route is compatible with the repository’s GitHub Pages export model and adds no dependency or server requirement.
- Mapping is conservative: profile/work/education/course/contact/skill data is reused; unsupported empty sections, synthetic proficiency labels, an inferred country code, and a misleading build-time `lastModified` are omitted.
- Rich HTML résumé summaries retain their links in the page while the JSON consumer receives readable plain text.
- The visible JSON affordance follows the existing RSS-chip treatment, stays out of print, and remained aligned at the 320px layout.
- Keyword ordering is deterministic with an explicit English locale when competency values tie.

I independently validated the emitted artifact with Draft 4 plus format checking against the official pinned [JSON Resume v1.0.0 schema](https://raw.githubusercontent.com/jsonresume/resume-schema/v1.0.0/schema.json): valid. I also checked the high-salience Promptfoo résumé claims against the [OpenAI acquisition announcement](https://openai.com/index/openai-to-acquire-promptfoo/) and [Promptfoo’s announcement](https://www.promptfoo.dev/blog/promptfoo-joining-openai/); the acquisition and published adoption figures represented in the shared résumé data are supported.

### Verification on this exact head

- `npm run format:check` — clean
- `npm run lint` — 182 files clean
- `npm run type-check` — clean
- `npm run og:check` — clean
- `npm test` — 60 files, 489 tests passed
- `npm run build` — 17 static routes, including `/resume.json`
- `npm run verify-export` — 14 pages plus sitemap, feed, and résumé validation passed
- Official-schema validation — valid
- Static HTTP check — `Content-Type: application/json`, 10,599 bytes
- Browser QA at 320px — no overflow, console errors, or page errors
- Hosted checks — Cloudflare, code quality, tests, Ubuntu Node 22/26, macOS Node 26, Windows Node 26, and aggregate CI all passed

No additional inline comments are warranted on this head.

PR #941, opened 2026-07-28
fix(a11y): strengthen contrast, focus, and forced-colors support

mldangelo:
## Why

The site had no explicit forced-colors treatment, reduced-motion hid a static paper texture, and focus-ring decisions were split between semantic tokens and direct palette references. The current-role amber also sat below the design’s intended 3:1 non-text floor on both light surfaces.

This is an accessibility and resilience improvement, not a blanket WCAG-conformance claim. The résumé already states `Present` in visible text, so the redundant timeline dot is not required to meet SC 1.4.11; increasing its contrast is still a useful visual-design choice. [W3C guidance for non-text contrast](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html)

### Contrast and focus

- Changes the light `--color-signal-mark` from `#e8930c` to `#bc770a`, moving it from 2.36:1 / 2.16:1 to 3.51:1 / 3.21:1 against the two paper surfaces.
- Adds `--color-focus-ring` for page-surface controls and `--color-focus-ring-fill` for accent-filled controls.
- Routes every explicit focus `outline` and `outline-color` declaration through those semantic tokens, including the skip link, hero résumé link, slide-menu controls, mobile links, and résumé navigation.
- Keeps the existing 3:1 control-boundary floor as a deliberate design constraint without claiming every visibly labelled border is independently required by SC 1.4.11.
- Adds a dependency-free opaque-hex contrast helper using the current WCAG 2.2 sRGB breakpoint (`0.04045`). Alpha-bearing colors are rejected because they cannot be scored without an explicit backdrop. [WCAG 2.2 luminance definition](https://www.w3.org/TR/WCAG22/#dfn-relative-luminance)

### Forced colors

A dedicated mode override now:

- restores real text underlines where gradient underlines disappear;
- uses system color pairs for button and selected-filter states;
- keeps the current résumé marker filled rather than letting it collapse into the hollow completed markers;
- preserves current-location cues in desktop navigation, the mobile menu, and the résumé section nav;
- removes portrait blend/filter treatments in normal, hover, dark, and footer contexts;
- removes the fixed paper-grain layer explicitly.

That last rule is intentional: forced colors removes non-URL background images, but URL-backed images can remain. The grain is a data URL, so it cannot be assumed to disappear automatically. [CSS Color Adjustment specification](https://www.w3.org/TR/css-color-adjust-1/#forced)

### Reduced motion

The reduced-motion media query still collapses animation and transition durations, but it no longer hides the static paper grain. The rationale is scoped to motion; it no longer makes unportable claims about compositor promotion or guaranteed repaint behavior.

## Regression coverage

- Contrast tests read the actual light/dark token sources and pin the relevant ratios.
- A repository-wide source check rejects explicit outline colors that bypass the two focus-ring tokens.
- Forced-colors tests pin import order, desktop/resume active bars, the mobile non-color cue, and higher-specificity portrait states.
- Browser QA verifies the emitted stylesheet rather than assuming source-order behavior.

## Verification

- `npm run format:check`
- `npm run lint` — 181 files
- `npm run type-check`
- `npm test` — 472 tests in 59 files
- `npm run og:check`
- `npm run build` — 16 static routes
- `npm run verify-export` — 14 pages
- Chromium QA at 320 px and 1280 px in forced-colors, reduced-motion, light, and dark modes: active desktop/resume/mobile navigation, filled filter state, current-role marker, hero/footer/hover portraits, paragraph-link underline, static grain behavior, keyboard focus, console errors, and horizontal overflow


mldangelo:

src/lib/contrast.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  const expanded =
+    body.length === 3 || body.length === 4
+      ? body
+          .split('')
+          .map((c) => c + c)
+          .join('')
+      : body;
+
+  if (!/^[0-9a-fA-F]{6}([0-9a-fA-F]{2})?$/.test(expanded)) {
```
mldangelo: Please reject alpha-bearing input here (except perhaps an explicit `ff`) or accept a backdrop and composite it. Right now `contrastRatio('#00000000', '#ffffff')` returns 21:1 because the alpha byte is discarded, even though the foreground is fully transparent. That contradicts the helper’s “opaque” contract and can let a future 4/8-digit token falsely pass the build-time floor; the new test currently enshrines the unsafe behavior.

app/styles/tokens/colors.css
```diff
@@ -33,15 +33,45 @@
...
   /* ===================
      SIGNAL — amber
      Reserved for live/measured values and the
      current-position marker. Never decorative.
-     --color-signal      text-safe on page bg (AA)
+     --color-signal      text-safe on page bg (5.5:1 raised, 5.0:1 page)
      --color-signal-mark graphic use only (ticks, dots, rules)
+
+     The mark is not decoration: the filled dot on the resume spine is the
+     only thing that says a role is current, so SC 1.4.11 applies and it owes
```
mldangelo: This is not the only indication that a role is current: the same experience row prints a date range ending in `PRESENT`. Improving the dot’s contrast is still worthwhile, but please describe it as the visual/spine indicator that reinforces the textual state. The current wording overstates both the accessibility dependency and “nobody could see” claim.

mldangelo (commented):
Verdict: **changes recommended before merge**.

The visible work is directionally strong and the forced-colors implementation held up in a real Chromium forced-colors session: hero prose links regain an underline, the portrait drops its blend/filter, the current-role marker remains filled, active skill filters use system highlight colors, and reduced-motion no longer removes the static paper texture. Keyboard traversal also showed the new filled-control ring in both light and dark themes.

I found one correctness hole in the new contrast “floor”: the exported parser accepts alpha-bearing hex colors and silently discards alpha, so a transparent token can receive an opaque contrast score. The current tokens are opaque, so this is not a current rendered regression, but it makes the new guard capable of false passes and its test currently locks in that behavior. Please reject non-opaque alpha (or require a backdrop and composite it) before treating this helper as a reusable token gate.

The amber adjustment itself is sensible, but the rationale should be factual: the date range already says “PRESENT,” so the dot reinforces current status rather than being its only expression.

- Prettier check and Biome check: clean (180 files)
- TypeScript: clean
- Vitest: 58 files / 466 tests passed
- Export verifier: 14 pages passed
- Real browser QA: forced colors, reduced motion, light/dark keyboard focus, active filter state, and horizontal overflow checked at 1280px

mldangelo:
Implemented in #960.

The header said the utility is "for opaque sRGB hex colours" while `parseHexColor` accepts 4- and 8-digit hex and ignores alpha — and a comment further down the same file already said so, so the file contradicted itself. The header now states that an alpha channel is accepted and ignored, and why that makes it a floor rather than a proof.

Kept the existing caveat paragraph intact, since it is load-bearing: token-pair contrast is not rendered contrast, because the paper grain (`body::before`), the header's `backdrop-filter`, and the portrait's `mix-blend-mode: multiply` all sit between the two colours being compared.

mldangelo:

mldangelo (commented):
Fresh current-head review of `906f606215839faf7fcdf280d53dbe17b30f0f63`: **changes recommended before merge**.

The rendered accessibility work itself is solid: I checked the built resume at 1280px in dark mode and 320px in light mode, verified that the new amber marker and filled-control focus-ring tokens resolve as intended, exercised a filled skill-filter focus state (2px `#f2f4f7` ring against the dark blue fill), confirmed there is no horizontal overflow, and inspected the emitted forced-colors rules. The current role also visibly renders both the amber dot and `MARCH 2026 – PRESENT`.

One correctness problem remains unresolved on this exact head:

1. **The “opaque” contrast gate silently produces false passes for transparent colors.** `parseHexColor` accepts 4/8-digit hex and discards alpha, even though the module and `contrastRatio` both promise opaque colors. I executed the shipped helper directly: `contrastRatio('#00000000', '#ffffff')` returns `21`, although fully transparent black composites to white and therefore has 1:1 contrast on that backdrop. The current test deliberately locks in the unsafe behavior. The present callers happen to be opaque, but this helper exists as a reusable build-time gate, so accepting an input it cannot score correctly defeats the gate's purpose. Please either reject alpha-bearing inputs or add an API that composites against an explicit backdrop, and make the tests pin that safe contract. Merely documenting that alpha is ignored does not prevent the false pass.

A smaller wording issue should also be corrected: the token comment and test say the dot is “the only thing” identifying a current role, but the rendered date explicitly says `PRESENT`. The dot is a useful redundant non-text indicator and still needs adequate contrast; the rationale should say that rather than make a factually false exclusivity claim.

- targeted contrast suite: 15/15 pass
- full Vitest suite: 466/466 pass across 58 files
- browser QA: 1280×900 dark and 320×800 light, focus state exercised, no overflow

mldangelo:

mldangelo (commented):
Final exact-head re-review at d6bea58.

Disposition: no remaining blocker. The contrast helper now enforces its opaque-sRGB contract: 4/8-digit alpha-bearing inputs are rejected instead of silently scoring transparent black as opaque black. The file-level and function-level documentation agrees with that behavior, and the status-dot rationale correctly describes it as reinforcement for the visible PRESENT text rather than the only current-role indicator.

All three prior review threads are addressed. Formatting, lint, type checking, 467 tests, build/export, and token contrast checks pass.

mldangelo (commented):
## Exact-head re-review — `8ab461558fe3f5539238e62be297f2c1bbda0dcd`

I reviewed the complete diff again, checked every existing thread, compared the claims with the current W3C specifications, inspected the emitted CSS, and exercised the relevant states in Chromium. Because this is our own PR, this is a `COMMENT` review rather than a self-approval.

### Findings from this pass

1. **[P1, fixed] The PR made a false WCAG-conformance claim.** The résumé visibly says `Present`, so the amber dot is redundant information. W3C’s SC 1.4.11 guidance excludes a graphic when visible text conveys the same information. Raising the dot from 2.36:1 / 2.16:1 to 3.51:1 / 3.21:1 remains a sound design improvement; it is now documented as a chosen floor rather than a mandatory pass/fail result. The same distinction now applies to visibly labelled control borders. [W3C non-text contrast guidance](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html)
2. **[P1, fixed] Forced colors lost two current-location indicators.** The original override preserved the desktop nav bar only. Mobile navigation used color alone, and the résumé section-nav pseudo-element flattened into the Canvas. The exact head now gives the active mobile item an underline and maps both desktop and résumé bars to `LinkText`; source tests pin all three paths.
3. **[P1, fixed] Portrait resets lost to specificity.** Import order did not beat `.site-footer-new .footer-avatar img` or the light/dark hero hover selectors. The footer remained grayscale and hover could restore filters in forced colors. The override now matches those contexts explicitly, and Chromium confirms normal, footer, hover, and dark portrait filters all resolve to `none` with normal blending.
4. **[P2, fixed] The background-image rationale contradicted the CSS Color Adjustment specification.** Forced colors removes non-URL background images, not every background image. The paper grain is a data URL and therefore can remain; `body::before { display: none }` is necessary, but the prior explanation was wrong. The comments and contributor guidance now make that boundary explicit. [CSS Color Adjustment specification](https://www.w3.org/TR/css-color-adjust-1/#forced)
5. **[P2, fixed] “Every ring is one decision” was not true.** Five focus outlines still bypassed the new semantic tokens in the skip link, hero résumé link, slide-menu close button, mobile menu, and résumé nav. Those are now tokenized, and a repository-wide test rejects any non-`none` outline declaration that does not use `--color-focus-ring` or `--color-focus-ring-fill`.
6. **[P2, fixed] The contrast helper used the superseded luminance breakpoint.** `0.03928` is now `0.04045`, matching the current WCAG 2.2 sRGB definition. W3C notes that this has no practical effect for 8-bit channels, but code that claims to implement the standard should use the current formula. [WCAG 2.2 relative luminance](https://www.w3.org/TR/WCAG22/#dfn-relative-luminance)
7. **[P2, fixed] Several implementation claims were too strong.** Vitest checks these tokens at test/CI time, not during `npm run build`; `Highlight`/`HighlightText` is the semantic selection pair, not a guaranteed contrast against every user Canvas; and `contain: strict` plus `translateZ(0)` does not normatively guarantee compositor promotion or zero repainting. The revised prose states only what the code and specifications establish.
8. **[P3, fixed] One forced-colors selector was stale.** The log hover marker already uses ink rather than amber and did not belong under the “amber/state only a colour” rationale. The redundant override is gone.

### Design and necessity

This PR is necessary. Forced-colors users previously lost gradient link lines and selected-state distinctions, while reduced-motion users lost a static texture unrelated to motion. The dedicated mode override is the right architecture: it remains after theme CSS and before print CSS, uses user-controlled system colors, and limits itself to information the forced palette cannot recover automatically. The new contrast helper is appropriately small and fail-closed for alpha; its documentation correctly describes token tests as a floor, not rendered conformance.

### Verification

- Format, Biome lint (**181 files**), TypeScript, and OG asset checks pass.
- Vitest passes: **472 tests in 59 files**.
- Production build exports **16 static routes**; `verify-export` validates **14 pages**.
- Chromium QA at 320 px and 1280 px covers forced colors, reduced motion, light/dark themes, selected filters, current markers, desktop/resume/mobile active navigation, normal/hover/footer portraits, paragraph-link underlines, static grain behavior, keyboard focus, and horizontal overflow.
- Reduced-motion keeps `body::before` visible at opacity `0.022` while transition duration collapses to `0.01ms`.
- Forced-colors hides the data-URL grain explicitly, preserves the active résumé bar as a 2 px `LinkText` rule, underlines the active mobile route, and leaves hero/footer portraits unfiltered.
- The browser console has no errors; the only warnings are Next font-preload notices from serving the export through the local static test server.
- Hosted exact-head checks all pass: Cloudflare Pages, dependency installation, code quality, test, Ubuntu 22/26, macOS 26, Windows 26, and CI Success.

### Verdict

No remaining blocking finding on this exact head. The implementation improves real accessibility behavior, and the PR now describes that improvement without overstating WCAG requirements or browser guarantees.

PR #940, opened 2026-07-28
feat(projects): publish a sourced register of selected work

mldangelo:
## Why

The projects page was hard to reach and a decade out of date: `/projects` was absent from primary navigation, while its visible entries stopped in 2015. The old set was also broader than “four student hackathons” — it included formal university satellite work — so this PR now describes the page without diminishing that history.

## What changed

- Makes `/projects` a primary navigation destination.
- Recasts the page as **Selected work** plus an **Archive** of earlier student projects, hackathons, and experiments.
- Adds current production work from the résumé while keeping older projects rather than deleting them.
- Uses month-precision date ranges and an explicit `ongoing` flag, so a missing end date is not silently treated as “Present.”
- Makes images optional and keeps entries without a public destination visibly inert with “No public link.”
- Generates complete page metadata through the shared metadata helpers.
- Uses ordered-list/list-item semantics for both collections and adds units to the visible counts (`5 projects`, not a context-free `5`).

## Accuracy and source discipline

Every selected-work entry now declares a `sourceWork` résumé role. Tests verify its dates and complete numeric tokens against that specific role rather than searching the whole résumé or accepting substring matches (`20` no longer matches `2026`). Date validation also rejects impossible or non-ISO calendar dates.

The copy and destinations received a fresh live-source pass:

- **Codex Security:** described as work contributed to at OpenAI, not as something necessarily launched by the author. The public research preview predates the recorded OpenAI role start by three days. [OpenAI announcement](https://openai.com/index/codex-security-now-in-research-preview/)
- **Promptfoo:** adoption figures and the OpenAI acquisition remain supported by Promptfoo’s own announcement. [Promptfoo announcement](https://www.promptfoo.dev/blog/openai-acquires-promptfoo/)
- **Smile ID:** points directly to its current canonical domain, and the latency claim now says “7 seconds” instead of leaving the unit implicit. [Smile ID](https://smile.id/)
- **Arthena:** the compromised legacy domain/PDF path was removed after it redirected through spam and gambling pages. The project now points to the stable YC company record, and the prose separates the 2014–2022 operating role from Masterworks’ later 2023 acquisition. [Y Combinator profile](https://www.ycombinator.com/companies/arthena), [Arthena acquisition post](https://www.linkedin.com/posts/arthena_arthena-press-release-activity-7042233255940497408-EmBO)
[...]

PR #942, opened 2026-07-28
feat(writing): generate verified share cards for published posts

mldangelo:
## Summary

- Generate a dedicated 1200×630 share card for every published post, so article links use post-specific Open Graph and Twitter previews instead of the generic site card.
- Preserve each post's representative article image and measured dimensions in JSON-LD; the share card and article image remain deliberately separate concepts.
- Derive the card styling and readouts from source-controlled design tokens, post content, profile facts, checksum-pinned fonts, and locked renderer packages.
- Keep drafts out of committed cards and the static export. Development-only draft routes inherit `/og.png`, emit `noindex, nofollow`, omit canonical/publication metadata, and tolerate intentionally absent root-relative draft images without weakening published-post validation.

## Verification and safeguards

- Record every card digest and its material inputs in `public/og.meta.json`; `npm run og:check` validates the ledger and performs a fresh byte-for-byte render comparison.
- Preflight title and description fit, reject copy that cannot fit, and use `break-word` for unusually long tokens.
- Report CommonMark-aware prose word counts and **unique** external-link destinations. The three post cards now report 406 / 868 / 1,924 words and 9 / 9 / 17 unique external links.
- Render images before mutation, remove stale cards, and write the ledger last as the generation commit marker.
- Parse draft Markdown with the renderer's parser rather than filename heuristics: inline/reference/relative resources, passive raw HTML, `srcset`, same-origin absolute URLs, attachments, and frontmatter images are checked against the actual export.
- Fail closed for active or uninspectable raw HTML/CSS/data-document forms, including escaped CSS URLs and SVG/SMIL fetches.
- Remove three draft-only screenshots from `public/` (116,828 bytes total) while retaining safe fallback geometry in local draft previews.
- Keep published posts strict: missing, malformed, unreadable, relative, or traversal-based local images still stop the build.

## Validation

- `npm run format`
- `npm run lint` — 184 files clean
- `npm run type-check`
- `npm test` — 59 files, 537 tests passed
- `npm run og:check` — all four committed cards match a fresh render
- `npm run build` — 16 static routes generated; the draft route was not exported
- `npm run verify-export` — 14 pages passed, including draft-route and referenced-asset privacy checks
- Development route QA — the real draft returned HTTP 200, emitted `noindex, nofollow`, inherited `/og.png`, omitted its draft card path, and rendered all three absent screenshots with a 1200×675 fallback ratio
- Native card QA — all cards are opaque 1200×630 sRGB PNGs with no clipping, collisions, broken glyphs, or missing rules; every SHA-256 matches the ledger

The 1200×630 dimensions and explicit image dimensions/alt text align with the current [Next.js Open Graph image guidance](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/opengraph-image) and [Open Graph structured image properties](https://ogp.me/).


mldangelo:

scripts/generate-og.mjs
```diff
@@ -134,89 +174,238 @@ function card() {
...
+      typeset(post.description),
     ),
   );
 }
 
-const [display, mono] = await Promise.all([
+// The three type roles the site uses, fetched once and reused for every card.
+// `loadGoogleFont` is the only network dependency in this script, so one round
+// trip per card would be both slow and one more chance of a partial failure.
+const [display, mono, serif] = await Promise.all([
```
mldangelo: These remote font bytes are an unrecorded rendering input. `generatorDigest` hashes the three script sources and profile snapshot, while the ledger separately records posts/colors—but neither records the Google CSS response, resolved font URLs, or downloaded bytes. The same repository state can therefore render different PNGs without any declared input changing. This PR also adds a third remote face and expands the output from one card to four. Please use the repository’s version-pinned/self-hosted faces (or vendor compatible font files) and hash the exact bytes into the generator inputs.

scripts/og-inputs.mjs
```diff
@@ -0,0 +1,305 @@
...
+    file.endsWith('.md'),
+  );
+  const cards = [];
+
+  for (const file of files.sort()) {
+    const { data, content } = matter(
+      await readFile(join(directory, file), 'utf8'),
+    );
+
+    if (data.draft === true) continue;
```
mldangelo: This is a second, weaker publication boundary. `validatePostFrontmatter` rejects a non-boolean `draft` and invalid calendar dates, while this path treats `draft: "true"` as published and accepts dates such as `2026-99-99`. Full tests/build will eventually fail through the application reader, so deployment is mitigated, but `npm run og` can first write and expose a card in the public repository. Please share the validator/predicate (or extract a plain-JS boundary module) instead of relying on set-equality tests to keep two parsers synchronized.

scripts/og-inputs.mjs
```diff
@@ -0,0 +1,305 @@
...
+ * These are counted from the post itself for the same reason `/stats` counts
+ * its own source: a number typed into a card is a number that goes stale
+ * without anything failing.
+ */
+export function measurePost(markdown) {
+  const words = countProseWords(markdown);
+
+  return [
+    { label: 'Words', value: words.toLocaleString('en-US') },
+    { label: 'References', value: String(countReferences(markdown)) },
```
mldangelo: The visible label should match the measurement. `countReferences` de-duplicates exact outbound URL strings: two pages on one source count twice, and URL variants/anchors can count separately. That is “Outbound links,” not necessarily references or distinct sources. Renaming the label/function would keep the card’s factual readout defensible.

mldangelo (commented):

Per-post share cards are a real improvement: the current homepage card is not a useful preview for an essay, and the three post cards are visually coherent, legible, correctly sized, and wired into both OpenGraph and Twitter metadata. The structured-data split between a representative article image and a 1.91:1 share card also makes sense. I inspected all four committed 1200×630 PNGs at full resolution, checked the generated HTML, confirmed the draft slug is absent from the card directory/ledger/export, and verified the full artifact gates.

The blocker is reproducibility. The new ledger is presented as a digest of every input that determines the pixels, but the cards still resolve three font files dynamically through Google’s CSS endpoint, and neither the resulting URLs nor font bytes enter `generatorDigest`. That means identical source, profile, colors, and post content can produce different PNG bytes when the remote response changes. It also conflicts with this repository’s existing “variable and self-hosted” font rule and makes `npm run og` require the network even though the exact site faces are already version-pinned through Fontsource. Please render from versioned local/vendor font assets and include those bytes (or immutable asset digests) in the input ledger.

Two smaller integrity/wording issues are inline:
- the script has a second, weaker interpretation of frontmatter/draft status; the application validator prevents a bad deployment, but the generator can write an artifact first;
- “References” is not what is being measured—the code counts distinct outbound URL strings.

- Prettier and Biome: clean (179 files)
- TypeScript: clean
- Vitest: 58 files / 468 tests passed
- `og:check`: four 1200×630 cards current
- Export verifier: 14 pages passed, including draft assets and local share images
- Artifact QA: all four PNGs opened at original resolution; post-card fit, typography, dates, descriptions, and readouts checked; emitted OG/Twitter/JSON-LD paths inspected

mldangelo:
Thanks — I checked this one and it does not reproduce, so I have not changed the test.

The claim is that `path.join()` discards preceding segments when it meets a leading `/`. That is `path.resolve()`, not `path.join()`. Verified on this Node:

```
$ node -e "const p=require(\"path\");
console.log(p.join(\"/repo\",\"public\",\"/images/x.png\"));
console.log(p.resolve(\"/repo\",\"public\",\"/images/x.png\"));"
/repo/public/images/x.png     <- path.join, correct
/images/x.png                 <- path.resolve, what the review describes
```

`path.join` concatenates and normalises; only `path.resolve` treats a leading `/` as an absolute reset. So the assertion checks `<repo>/public/...` as intended and passes on POSIX runners, which matches CI being green on this file.

There *is* a related platform issue in this family, and it was real: comparing an OS-specific `path.join` result against a forward-slash manifest key breaks on Windows. That one is fixed separately in #959 with `path.posix.join`. It does not apply here because nothing compares this path against a URL-style key.

Leaving as-is. Happy to revisit if you can show a platform where `path.join` behaves as described.

mldangelo:
Beyond the `path.join` note (replied separately — it does not reproduce), the draft-leak risk on this PR was verified end to end rather than assumed.

The built export contains exactly three cards, one per *published* post. The `draft: true` post has none, so the leak class this PR could have introduced does not occur.

A related gap was found afterwards and is being fixed separately: the draft-asset gate matches on filenames, so it catches a card named after a draft slug but *not* a draft's article images, which live in `public/images/writing/codex-desktop-app-post/` — a directory named after neither the slug nor anything the gate matches. Three unpublished screenshots ship while the gate reports "14 pages OK". That exposure is pre-existing on `main` and the earlier broader scan would not have caught it either, but the gate should. In flight, gating on frontmatter rather than filenames.

mldangelo:

mldangelo (commented):
Fresh current-head review of `398438d9564cdacf337786af9143d7cfab7c93b0`: **changes recommended before merge**.

The feature is worthwhile: a post-specific card is materially better than every essay sharing the homepage image. I inspected all four committed PNGs at their native 1200×630 size; the typography, hierarchy, line wrapping, and readout spacing are clean, including the long Claude Code title. The built HTML for all three published posts points both Open Graph and Twitter metadata at the correct per-post image with 1200×630 dimensions and an alt value, and the current real draft produces neither a route nor a card.

Two pipeline-integrity issues remain on this exact head:

1. **The generator is not deterministic even though the script and documentation say it is.** The three actual font byte streams come from unversioned `fonts.googleapis.com` CSS requests at generation time. Those bytes are not repository inputs and are not included in `generatorDigest`; that digest covers three script sources plus the profile snapshot. `og:check` also does not render the expected image—it only checks that the committed PNG still matches the digest recorded beside it. Consequently an upstream font revision can make the same commit generate different binaries, while CI continues to call the old binaries “generator current.” That creates unexplained binary churn the next time anyone runs `npm run og`, and it contradicts the stated reproducibility rationale. Please render from versioned repository/package font assets (the site already pins the three Fontsource families) and include the exact font bytes or their hashes among the generator inputs.

2. **The artifact reader has a weaker frontmatter contract than the application reader at the exact boundary where draft isolation matters.** `readPostCards` skips only `data.draft === true`, so `draft: 'true'` is treated as published; its date check accepts any digit-shaped value such as `2026-99-99`, which formats as `Invalid Date`. I executed the application validator against those inputs and it correctly rejects both, while the generator control flow includes the non-boolean draft and accepts the date shape. Running `npm run og` can therefore write a public PNG and ledger entry for content the subsequent site build refuses. Please share the canonical validation logic rather than maintaining a second, looser parser, and add malformed-draft and impossible-date fixtures to the generator tests.

Wording/design nit: `References` overstates what is measured. The implementation counts distinct exact inline HTTP(S) URL strings; two URLs from one publisher count twice, duplicate links collapse, and reference-style/HTML links are invisible. `Outbound links` (or a deliberately parsed citation count) would describe the value more honestly.

- focused OG/metadata/export tests: 68/68 pass
- full Vitest suite: 468/468 pass across 58 files
- `og:check`: pass for 4 cards
- original-resolution visual inspection: all 4 PNGs clean
- exported metadata and the exact three post-card files verified

mldangelo:

mldangelo (commented):
Final exact-head re-review at 7139e37.

Disposition: no remaining blocker. Generated post cards are reproducible: all three exact versioned font URLs and SHA-256 digests are recorded and verified before rendering. The app, generator, and export verifier share one strict frontmatter validator, so malformed dates or non-boolean draft flags cannot cross a weaker publication boundary. The card readout now says “External links,” matching the URL count it actually computes, and output ordering is deterministic.

The Copilot path.join concern is a false positive for the same Node semantic noted on #940; join preserves the public prefix here, while resolve would reset it. All four threads are closed. Formatting, lint, type checking, 477 tests, Open Graph determinism, build/export, and card visual review pass.

mldangelo (commented):
Re-review of exact head `93589ad36409a0921c834511b6571e8725729688` (submitted as COMMENT because this is an author-owned PR).

I reviewed the complete diff, generated PNGs, metadata behavior, failure modes, documentation, and hosted results rather than treating earlier bot reviews as coverage. The feature is worthwhile at this head: each published article gets a specific link-preview card, while JSON-LD continues to describe the article's actual representative image. The generated binaries are now tied to reviewable source inputs and a reproducible renderer instead of being opaque assets.

Findings fixed in this pass:

- Corrected the external-link measurement to include linked-image Markdown. "Shipping with Claude Code" now defensibly reports 17 distinct outbound URLs rather than 16.
- Replaced the broad angle-bracket stripping in prose measurement with recognized HTML-tag removal, so comparison prose such as `latency < 50ms and throughput > 1k` is not silently deleted.
- Added the locked Next, React, and Sharp package versions and integrity hashes to the renderer fingerprint. A lockfile renderer change now fails verification even when scripts and post inputs are unchanged.
- Strengthened `npm run og:check` from ledger-only validation to a fresh byte-for-byte rerender comparison. I fault-injected a modified PNG and matching forged ledger digest: the structural check passed, while the fresh-render gate correctly failed.
- Added fit preflight logic plus `break-word` handling for long tokens. Tests cover long vertical copy and unbreakable text; an overlong injected title fails closed and names the affected slug.
- Bounded remote font retrieval to two 15-second attempts, while keeping checksum mismatches immediate and explicit.
- Made generation recoverable by rendering all cards before mutating outputs and writing the ledger last as the commit marker.
- Extended draft leakage verification to HTML and other exported files named after a real draft slug, including noindex HTML outside the normal writing route.
- Updated the documentation and workflow wording, including correcting the former claim that Sharp was never imported in this repository.

Validation on this exact head:

- Full local suite: 58 files and 483 tests passed.
- Lint: 181 files passed; type checking and format verification passed.
- Production export: 16 routes built; all 14 exported pages passed verification.
- Node 26.5 reproduced all four committed cards byte-for-byte.
- Browser QA confirmed the post canonical URL, per-post Open Graph and Twitter metadata, 1200x630 card dimensions, and the distinct JSON-LD article-image metadata.
- Original-resolution inspection of all four generated cards found no clipping, overflow, illegible copy, or unintended typography regression.
- Hosted checks passed on macOS Node 26, Ubuntu Node 22 and 26, Windows Node 26, tests, code quality, and Cloudflare Pages.
- All four earlier inline review threads are resolved.

I found no remaining blocking correctness, accessibility, content-accuracy, or design issue at this exact head.

mldangelo (commented):
Re-review completed against exact head `378397da2194bd7856a4cd94607ba7d41f8562d7`.

This PR is now justified and materially improves the site: published essays get specific, reproducible share previews instead of reusing the homepage card, while JSON-LD retains the actual representative article image. The 1200×630 output, explicit width/height, and useful alt text are consistent with the current [Next.js Open Graph image guidance](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/opengraph-image) and [Open Graph structured image properties](https://ogp.me/).

The original head was not ready to merge. This pass reproduced and fixed several concrete blockers:

1. **The real development draft route returned HTTP 500.** Metadata generation unconditionally measured `/og/writing/<draft-slug>.png`, even though the generator correctly refuses to create draft cards. Draft metadata now inherits `/og.png`, emits `noindex, nofollow`, and omits canonical, `og:url`, author, and publication-time fields.
2. **Draft-asset privacy verification was fail-open.** The old export gate only compared exported path segments with the draft slug. The actual draft referenced three public WebPs under an unrelated directory, so all 116,828 bytes shipped while `verify-export` still said “14 pages OK.” Those files are now removed from `public/`, and the verifier checks resources actually declared by draft content.
3. **The first replacement scanner had meaningful syntax gaps.** A shared parser-backed boundary now covers inline, full/collapsed/shortcut reference, escaped/balanced, nested, relative, same-origin absolute, attachment, frontmatter, passive raw-HTML, and every `srcset` candidate form. It ignores code and resolves local targets from the eventual post route/base path.
4. **Adversarial review found two additional bypass classes.** Escaped CSS `u\72l(...)` and SVG/SMIL resource mutation could hide a fetch from simple attribute inspection. Raw style attributes and SVG/active resource containers now fail closed, with exact regression cases. Active data documents are rejected; raster data images remain allowed.
5. **Draft tolerance had to remain narrowly scoped.** Development drafts may reserve 1200×675 geometry only when a root-relative local image is genuinely absent. Published posts still fail for missing, malformed, unreadable, relative, or traversal-based files.
6. **Three card readouts overstated prose length.** Ordered-list markers were counted as words. Independent counting now agrees at 406, 868, and 1,924 words. The link metric is explicitly labeled “Unique external links,” matching its 9 / 9 / 17 unique-destination calculation rather than implying occurrence counts.
7. **Several wording claims were inaccurate.** “The one asset” now describes the full card set; the generator no longer claims published posts never change; and the adapting guide correctly scopes regeneration to published-post changes.

Visual QA found all four cards to be opaque 1200×630 sRGB PNGs with no clipping, collision, missing rule, or broken glyph. The ledger hashes reconcile with every committed image, and a fresh render matched byte for byte.

Validation on the repaired head:

- `npm run format`
- `npm run lint` — 184 files clean
- `npm run type-check`
- `npm test` — 59 files / 537 tests passed
- `npm run og:check` — four ledger checks and four fresh-render matches
- `npm run build` — 16 static routes; the draft was not exported
- `npm run verify-export` — 14 pages passed
- Real development draft route — HTTP 200, `noindex, nofollow`, `/og.png`, no draft-card path, three safe fallback image slots

The build-time GitHub statistics request used the repository’s existing fallback because the sandbox could not resolve `api.github.com`; compilation, TypeScript, static generation, and export verification all completed successfully, so that was unrelated to this PR’s share-card path.

I found no remaining actionable blocker after the final independent adversarial pass. All four pre-existing inline threads are resolved. Because this is our own PR, this is intentionally a `COMMENT` review rather than an approval.

PR #943, opened 2026-07-28
fix(writing): improve figures, code fences, and heading anchors

mldangelo:
## Summary

- Render standalone and linked article images as semantic figures, using the site's image-title convention for visible captions.
- Let measured local figures use the wider page measure without enlarging them past their intrinsic width.
- Show fenced-code languages visually and to screen readers, while adding keyboard focus only when a fence actually overflows.
- Normalize generated heading IDs, preserve compatible fragments from the old slug scheme, and avoid collisions with raw HTML IDs and footnote targets.

## Improvements from review

- Migrated the GitHub contributions qualification into the figure caption convention so a published post exercises the feature.
- Fixed suffix collisions such as two elements receiving `repeat-2`, including collisions with the `section` fallback.
- Preserved one working legacy alias when the old renderer emitted the same heading ID more than once.
- Kept footnote links and footer IDs on markdown-to-jsx's original slug scheme, including dotted labels.
- Re-measured fence overflow after font loading, viewport changes, and client-side content changes.
- Added explicit Left and Right arrow scrolling for focused, overflowing fences.
- Preserved safe attributes on raw HTML headings, trimmed whitespace-only captions, and constrained long language labels.

## Validation

- 479 tests passed across 57 files.
- Formatting, lint, and TypeScript checks passed.
- The production build completed with 16 routes.
- Export verification passed across 14 pages.
- Browser QA passed at 1280x900 and 375x812 in both themes.
- Desktop figures rendered at 832px and mobile figures at 327px without document overflow.
- Legacy fragments landed below the fixed header. On mobile, only the two genuinely overflowing fences entered the tab order, and ArrowRight scrolled the focused fence.


mldangelo:

src/components/Writing/PostContent.tsx
```diff
@@ -10,37 +19,149 @@ interface PostContentProps {
...
+        // No wrapper element: the default `<div>` around a multi-block document
+        // becomes the only child of `.prose`, so any per-block rule written as
+        // `.prose > *` silently lands on the wrapper and constrains everything
+        // inside it — which is exactly what stopped a wide figure widening.
+        wrapper: null,
+        // One slug scheme for the whole site. The default slugifier keeps every
+        // separator it deletes, so `--dangerously-skip-permissions` became
+        // `on-using---dangerously-skip-permissions` while every other page on
+        // the site derived `on-using-dangerously-skip-permissions`.
+        slugify: createHeadingId,
```
mldangelo: This changes a public fragment contract without a compatibility target. The existing renderer emits `on-using---dangerously-skip-permissions`; at this head that ID is absent and only `on-using-dangerously-skip-permissions` exists. A repo search cannot find external bookmarks/shared links, and the adjacent AGENTS rule explicitly says to preserve stable heading IDs. Please render legacy aliases for every changed published ID (with matching scroll margin) while using the new canonical IDs. Also test duplicate headings: this callback independently returns the same ID for equal headings, while the repository already has `createUniqueHeadingIds` for that case.

src/components/Writing/PostContent.tsx
```diff
@@ -49,14 +170,39 @@ export default function PostContent({
...
+                    // annotation only and must not be announced twice.
+                    <span className="prose-fence-lang" aria-hidden="true">
+                      {language}
+                    </span>
+                  ) : null}
+                  {/* The fence has always scrolled horizontally, but only for a
+                      mouse. A named, focusable region gives the keyboard the
+                      same reach. */}
+                  <pre
+                    tabIndex={0}
```
mldangelo: Please do not add every fence to the tab order unconditionally. Measured on the real shipping post: all 4 fences have `scrollWidth === clientWidth` at 1280px, and only 2 of 4 overflow at 375px, but all four become focus stops. Detect actual overflow (including resize changes) and set `tabIndex=0` only then. The unconditional `role="region"` also creates four landmarks—two with the same “markdown code block” label—despite W3C guidance that regions represent sufficiently important navigable sections and repeated regions need unique labels: https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/

src/components/Writing/PostContent.tsx
```diff
@@ -10,37 +19,149 @@ interface PostContentProps {
...
+const LANGUAGE_PREFIX = 'language-';
 
 function isRootLocalImage(src: string): boolean {
   return src.startsWith('/') && !src.startsWith('//');
 }
 
+/**
+ * The image a block-level paragraph exists only to carry, if it carries one.
+ *
+ * A Markdown image's CommonMark title — `![alt](src "title")` — is its caption,
```
mldangelo: Please frame this as a site-specific authoring convention, not CommonMark semantics. CommonMark renders this token as the image’s `title` attribute, not a caption (``): https://spec.commonmark.org/spec#images. This branch intentionally repurposes and removes that title, and no current image in `content/writing/` uses the syntax. Either migrate an actual authored caption so the feature has a real consumer, or document the local convention and preserve the title if it remains useful.

app/styles/pages/writing.css
```diff
@@ -389,13 +400,51 @@
...
   padding: 1rem;
   overflow-x: auto;
-  margin: 1.5rem 0;
+  margin: 0;
+}
+
+/* The fence is keyboard-reachable so its horizontal scroll is too, which means
+   it needs a focus ring of its own. */
+.prose pre:focus-visible {
+  outline: 2px solid var(--color-accent);
```
mldangelo: This introduces another hard-coded focus-ring width while the design already has `--rule-strong`, and PR #941 is centralizing ring color as `--color-focus-ring`. At minimum use `var(--rule-strong)` here; if #941 lands first, rebase and use the focus token too so this new surface does not immediately recreate the drift that PR removes.

mldangelo (commented):

The figure/layout work is useful and measured correctly. On the real shipping post I saw 640px prose with 832px figures at 1280px, 327px prose/figures at 375px, and no document overflow at either size. The code-fence plates are restrained, and keyboard arrow scrolling works when a fence actually overflows (the mobile unlabeled fence moved from `scrollLeft=0` to `29` on ArrowRight).

Two accessibility/compatibility issues block this version:

1. The slugifier deliberately replaces fragment IDs that are already published. The current/default renderer emits `#on-using---dangerously-skip-permissions`; this branch removes that target and emits `#on-using-dangerously-skip-permissions` only. Repository search cannot prove that nobody has bookmarked or shared the old public URL, and AGENTS.md explicitly says to preserve stable heading IDs. Please retain legacy alias targets while adopting the canonical scheme. Also account for duplicate headings: `createHeadingId` is not uniqueness-aware even though `createUniqueHeadingIds` already exists.

2. Every `pre` is now a tab stop and named `region` landmark regardless of whether it scrolls. On the real post, **0/4** fences overflow at 1280px and only **2/4** overflow at 375px, yet keyboard users must traverse all four in both layouts. The page also exposes two landmarks with the identical “markdown code block” label. WAI-ARIA defines `region` for sufficiently important sections users should navigate to in a page summary and says repeated regions need unique labels; the W3C guidance also warns that landmark value diminishes as the count grows: https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/ and https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/. Please make keyboard focus conditional on actual overflow and avoid/uniquely label region landmarks.

The caption convention and focus-ring integration notes are inline. Neither current published post actually uses Markdown image-title syntax, so the caption feature is presently unexercised in real content; the existing graph caption remains a separate italic paragraph.

- Prettier and Biome: clean (177 files)
- TypeScript: clean
- Vitest: 57 files / 462 tests passed
- Export verifier: 14 pages passed
- Real browser QA at 1280×900 and 375×812: all three posts opened; figure/prose geometry, page overflow, fence labels, focus rings, tab order, arrow-key horizontal scrolling, and legacy/new fragment presence inspected

mldangelo:
Copilot's overview raised no actionable defect here.

For traceability, the three defects this PR fixes were each verified against the source first:
- `imageSize.ts` parses the CommonMark title and markdown-to-jsx forwards it, but `PostContent.tsx:32` destructured only `{ alt, src }`, so every caption the author wrote was silently discarded.
- The `$9,986.20` screenshot is 1117×812 and was rendered at 640px, with a legible dollar figure as its entire point.
- Every fence already shipped its language into the DOM and nothing read it; `content/writing/claude-code-outage.md:12-14` types `← crashes` inside the fence because the renderer offered no way to mark a line.

The slug-unification half had a consequence found later: it renamed four already-published `<h2 id>` values on `/writing/shipping-with-claude-code/`. Legacy anchors are preserved in #951 by deriving the old scheme from markdown-to-jsx's own default slugifier, so no hand-maintained list can drift.

mldangelo:

mldangelo (commented):
Fresh current-head review of `5095c72f8b5250b9d9e773fec53d0932107ed2cc`: **changes recommended before merge**.

The visual layout work is good. In the real built article, both figures expand to 832px against the 640px prose measure at 1280px, collapse cleanly to 327px at 375px, and never create page overflow. The language plate is restrained and consistent with the site's mono annotation role. The implementation also preserves valid figure/paragraph nesting.

There are two accessibility/compatibility problems I would fix before landing:

1. **This changes four already-published fragment IDs without preserving aliases.** On the current public rendering, these are:
   - `on-using---dangerously-skip-permissions` → `on-using-dangerously-skip-permissions`
   - `browser-automation---chrome-for-interaction-playwright-mcp-for-screenshots` → `browser-automation-chrome-for-interaction-playwright-mcp-for-screenshots`
   - `plan-mode--writing-plans-to-files` → `plan-mode-writing-plans-to-files`
   - `claudemd--agentsmd` → `claude-md-agents-md`

   I loaded the first legacy fragment in the built site: the old ID is absent and the page remains at the top while the replacement heading sits roughly 2,500px down-page. Existing bookmarks and shared deep links therefore break. Please preserve the legacy IDs as aliases (or retain them as canonical IDs) in this PR.

   The replacement slugifier also does not make IDs unique. Rendering two `## Repeat` headings produces two literal `<h2 id="repeat">` elements. The repository already has `createUniqueHeadingIds`; the Markdown path needs equivalent collision handling plus a regression test for repeated/normalization-colliding headings.

2. **Every fence becomes a landmark and tab stop, whether or not it scrolls.** The published Claude Code article has four fences. At 1280px none overflow, yet all four receive `tabIndex=0` and `role="region"`; at 375px only two of four overflow. The same page also exposes two indistinguishable `markdown code block` region labels. This adds unnecessary keyboard stops and pollutes screen-reader landmark navigation. The [W3C ARIA APG says a region is for content important enough to navigate to and repeated regions need unique labels](https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/#region). Please avoid making ordinary snippets landmarks; ideally make only genuinely scrollable fences keyboard-focusable (a resize-aware overflow check is possible because this is already a client component). If regions are retained, each must at least have a meaningful unique label.

Two smaller accuracy/maintenance points:

- CommonMark does **not** define an image title as a caption; its canonical example renders `foo` as an `img title="title"`, not `figcaption` ([CommonMark example 572](https://spec.commonmark.org/0.31.2/#example-572)). A site-specific “titles become visible captions” convention can be reasonable, but the code and AGENTS.md should call it that. No image anywhere under `content/writing/` currently has title syntax, so this portion of the PR does not yet add a visible caption; either document and use the convention on the actual figures or defer it.
- `.prose pre:focus-visible` duplicates the existing global focus rule with a slightly different offset. It will also bypass the focus-ring tokens introduced by #941 if both PRs land. Let the global rule own this so the two changes compose.

- focused PostContent suite: 14/14 pass
- full Vitest suite: 462/462 pass across 57 files
- browser QA: 1280×900 and 375×812, light mode, figures/code blocks/focus/fragment behavior inspected, no horizontal page overflow

mldangelo:

mldangelo (commented):
Final exact-head re-review at b536d5c.

Disposition: no remaining blocker. Published heading fragments keep compatibility aliases while new normalized IDs remain unique, so bookmarks do not silently break. Code blocks enter the tab order only when they actually overflow (and update on resize), avoiding gratuitous focus stops and duplicate landmarks. The renderer treats image titles as this site's explicit caption convention while preserving the title attribute, and the focus treatment uses shared design tokens.

All four prior threads are addressed. Formatting, lint, type checking, 468 tests, build/export, and desktop/mobile browser checks pass.

mldangelo (commented):
Exact-head re-review of `271c8c05da6f1a448a8cdb42746cdbbff07a6895`.

Disposition: no remaining blocker after this remediation pass. This is an author `COMMENT` review, not an approval.

The fresh pass found several problems that the previous exact-head review had missed:

- The heading allocator could emit duplicate IDs when an authored heading already matched a generated suffix. For example, `Repeat`, `Repeat`, and `Repeat 2` produced two `repeat-2` values. The same class of collision affected the `section` fallback.
- Replacing markdown-to-jsx's slugifier globally broke footnotes. A dotted label such as `[^foo.bar]` linked to `#foobar`, while the footer could be rewritten to `foo-bar`. A heading and footnote with the same source could also compete for one ID.
- Repeated headings lost their old public fragment entirely because compatibility aliases were suppressed whenever the legacy value appeared more than once.
- Generated headings did not reserve authored raw HTML IDs, and the raw-heading override discarded safe attributes such as `class` and `data-*`.
- Fence focusability could become stale after the final font loaded or retained content changed during client navigation. In browser QA, the native ArrowRight behavior also did not move the focused horizontal scroller.
- The visible language plate was hidden from assistive technology with no equivalent text, long language names could overflow, and no published post exercised the new caption convention.

The current head addresses each item. Heading IDs are allocated from a global claimed set, footnotes keep the renderer's original link/footer scheme, one safe legacy alias survives repeated headings, and raw IDs are reserved without stripping authored heading props. Code fences remeasure on font readiness, resize, observation, and child changes; only actual overflow adds a tab stop; Left and Right arrows now scroll explicitly. Language text has a screen-reader equivalent and a constrained visual plate. The published GitHub contributions image now uses the documented figure-caption convention, with the qualification phrased more precisely.

The figure work remains useful and visually sound. At 1280x900, the real figures render at 832px beside a 640px reading measure. At 375x812 they collapse to 327px. Both themes remain free of document-level horizontal overflow, linked figures are block-level, the caption wraps cleanly, and the old `#on-using---dangerously-skip-permissions` fragment lands at the intended heading below the fixed header.

Validation on this exact head:

- format check, Biome lint, TypeScript, and `git diff --check`: pass
- focused anchor and PostContent suites: 47 tests pass
- full Vitest suite: 479 tests pass across 57 files
- production static export: 16 routes
- export verifier: 14 pages
- OG source check: pass
- browser QA: 1280x900 and 375x812, light and dark themes
- GitHub Actions: code quality, tests, macOS/Ubuntu/Windows builds, CI success, and Cloudflare Pages all pass

All four earlier inline threads remain resolved. I found no new external-input or script-execution boundary in the final diff: the renderer still handles repository-authored Markdown through React and markdown-to-jsx, without introducing a new HTML injection path.

PR #944, opened 2026-07-28
feat(icons): refresh icons, manifest, and browser chrome

mldangelo:
## Summary

- Replace 29 unlinked pre-redesign favicon assets with seven reachable icon and manifest files generated from the current profile and design tokens.
- Use Next's static icon and manifest conventions so the export publishes a favicon, PNG tab icon, Apple touch icon, and maskable manifest icon.
- Keep browser chrome synchronized with the resolved site theme across manual choices, system changes for new visitors, and client navigation, with a light no-JavaScript fallback.
- Pin the generator to an exact font digest and Next renderer version so input changes require explicit regeneration.

## Improvements from review

- Reject invalid 5- and 7-digit hex values before generated art is written.
- Bind the ledger to both generator source files, the exact font bytes, shared profile and token inputs, and the installed Next renderer version.
- Require the ledger and manifest icon directory to contain exactly the outputs this generator owns.
- Extend export verification to check every icon and manifest link, each exported target, relative manifest icons, the start URL, the scope, and repository base-path containment.
- Correct comments about public asset discovery, favicon links, hashed URLs, maskable safe zones, and `minimal-ui`.
- Preserve existing stored theme choices. New visitors follow system changes until they use the site toggle; legacy stored values cannot be distinguished as implicit or explicit.

## Validation

- 506 tests passed across 62 files.
- Formatting, lint, TypeScript, and diff checks passed.
- Icon regeneration from the pinned TTF reproduced all seven binary hashes.
- The production build completed with 19 generated routes.
- Export verification passed across 14 HTML pages, including icon and manifest integrity.
- Browser QA passed in light and dark themes at 1280x720 and 375x812, including the manual toggle, client navigation, one synchronized `theme-color` tag, and no console warnings.
- The maskable monogram stays inside the standardized 40% radius safe circle.


mldangelo:

mldangelo (commented):
Verdict: changes requested before this branch merges on its own.

The cleanup itself is worthwhile. The old icon directory was genuinely unreachable, the token-driven generator removes a real source of design drift, the generated icons look crisp, and the static manifest/icon conventions survive the Pages export. I do not think the generator/test complexity is gratuitous given that it replaces hand-maintained binary assets and pins them to the design tokens.

One user-visible bug remains in this PR, though: `theme-color` follows the operating-system media query, while the site’s theme follows the user’s explicit `data-theme` choice. Those are independent states. In Chrome on this exact head, with the OS preference light, I clicked the site control to choose dark. The result was:

- `<html data-theme="dark">`
- computed body background `rgb(13, 16, 21)` (`#0d1015`)
- the matching `(prefers-color-scheme: light)` theme-color remained `#f2f1ec`

That leaves light browser chrome over a manually dark page. `app/__tests__/theme-color.test.ts` proves that the two token values are emitted, but it cannot prove that a manual theme selection updates the active metadata. The implementation needs to synchronize the meta value with the resolved site theme (and with later system-scheme changes while “system” is selected), or use an equivalent mechanism that observes `data-theme`. The follow-up already present in the integration branch appears to address this, but standalone #944 is still open and should not land without it.

There is also an existing Copilot inline comment about `path.join('public', icon.src)` making the generated-file assertion Windows-dependent. I agree with that nit; use URL/posix path semantics because manifest paths are URL paths.

Verification on `ca014af22addf527f77d16b0c3ca47d312900351`:

- focused token/icon/theme tests: 3 files, 29 tests passed
- `npm run build`: passed
- `npm run verify-export`: 14 pages OK
- generated ICO/PNG dimensions and manifest/icon links inspected
- real-browser light/dark toggle reproduced the metadata mismatch above

So: necessary and directionally strong, but not yet correct as an independently mergeable PR.

mldangelo:
Implemented in #959.

`path.posix.join` for the containment check, so the comparison no longer depends on the host path separator.

One correction to the framing rather than the fix: I initially described this as a real red build, and that overstates it. `.github/workflows/node.js.yml` does run a `windows-latest` leg, but it is the *build* job; Vitest runs only in the `test` job, which is `ubuntu-latest` with no OS matrix. So this could never have reddened CI — it fails `npm test` for anyone developing on Windows, which is arguably harder to diagnose because no other contributor can reproduce it. Fixed on the merits.

mldangelo:

mldangelo (commented):
Fresh re-review of the current head `ca014af22addf527f77d16b0c3ca47d312900351`: **changes are still needed before this branch can merge independently.**

The branch has not changed since my prior review, and I reproduced the same user-visible bug in a freshly built export. With the operating system in light mode, the page initially had `data-theme="light"`, body `#f2f1ec`, and the light `theme-color` active. After using the site's own control to choose dark, the page correctly changed to `data-theme="dark"` and body `#0d1015`, but the media query state did not change: the browser still selected the light `#f2f1ec` metadata. The implementation models only the OS preference; the product exposes an independent persisted user choice. Please synchronize the meta value with the resolved `data-theme` (and continue following later OS changes only while the user choice is “system”). The corresponding follow-up in the integration branch demonstrates that this is fixable; standalone #944 still omits it.

Two narrower correctness issues also remain:

- Copilot's unresolved inline thread is valid: `join('public', icon.src)` uses filesystem separators to compare URL-derived paths, so the assertion is Windows-dependent. Use `path.posix.join` or an explicit URL-path string.
- The new ledger claims to bind the committed art to the inputs that produced it, but the most important binary input—the font returned by two live Google Fonts requests—is absent from `generatorDigest`. Google can change the CSS URL or font bytes without a repository change, so identical committed source/profile/tokens can produce different icons and then pass once the freshly generated output hashes are written. Satori's own documentation confirms WOFF2 is unsupported, which explains why the existing Fontsource asset cannot be passed through directly, but it does not make an unpinned network input deterministic. Vendor a supported TTF/OTF/WOFF file or otherwise version/hash the exact downloaded font bytes.

The cleanup is still worthwhile: the deleted directory was unreachable, the replacement icons and manifest export correctly, and deriving colors/name from shared data removes real drift. The issue is not necessity; it is that one of the new visible behaviors is wrong and the reproducibility contract is stronger than the implementation.

Fresh verification:

- format check, Biome lint, TypeScript: passed
- Vitest: 60 files / 480 tests passed
- OG integrity check: passed
- production static build: passed (including icon and manifest routes)
- export verifier: 14 pages OK
- `git diff --check`: clean
- real Chrome QA reproduced the exact light-OS/manual-dark mismatch above

This review supersedes no code changes because there were none; it confirms the prior blocker against the still-current exact head.

mldangelo:

mldangelo (commented):
Final exact-head re-review at 04c10b5.

Disposition: no remaining blocker. The replacement icon set is reachable through Next's static conventions, its generator pins an exact versioned Bricolage Grotesque TTF plus SHA-256, and the ledger uses portable URL-shaped paths. The theme-color tag is script-owned, follows the resolved site theme, and survives client navigation instead of being recreated stale by React. Windows path behavior is explicitly tested with path.posix.

The prior portability thread is addressed. Formatting, lint, type checking, 490 tests, deterministic icon checks, production build/export, and browser navigation/theme QA pass.

mldangelo (commented):
Exact-head re-review of `336010d8d54bbe9026f5d3f3552fb32cdbc7a122`.

**Disposition:** I found no remaining code, design, accessibility, wording, or security blocker in the branch itself. I would still avoid treating the PR as deploy-ready while the external Cloudflare Pages check is red. GitHub exposes no provider log for that failure, while every repository-owned build passed, so there is currently no evidence-backed code fix to make for it.

### What this pass found and fixed

- The generator's color parser accepted invalid 5- and 7-digit hex values. It now validates the only CSS hex lengths the standard permits, validates the custom-property name before constructing a regular expression, and has focused rejection tests.
- The generated-art ledger omitted two inputs that can change pixels: the shared parser source and the installed Next renderer version. Both are now bound into the digest, alongside the exact font URL and SHA-256.
- The test trusted paths supplied by the ledger before constraining them. It now reads only a hard-coded output allowlist and requires the ledger to contain exactly those owned files.
- Renaming or removing a generated manifest icon could leave the old file silently exported. The test now requires an exact directory inventory and fails with a removal instruction when an obsolete file remains.
- The export verifier did not check this PR's central contract. It now validates every exported page's manifest, favicon, PNG icon, and Apple icon links; same-origin and base-path containment; file existence; manifest JSON; relative icon targets; `start_url`; and `scope`.
- Comments and the PR description made several inaccurate claims: that the deleted public assets were unreachable, that the favicon had no HTML link, that the manifest URL was content-hashed, that `minimal-ui` guarantees a URL bar, and that the old implementation remained media-scoped. Those statements are corrected, and the stale generated description has been replaced.

### Necessity and design

This PR is justified. It removes 29 unlinked pre-redesign assets, replaces the live icon that still used the retired accent, and gives the export a coherent favicon, tab icon, Apple touch icon, and manifest set derived from shared design/profile inputs. A committed deterministic generator is reasonable for binary artifacts that must remain synchronized with CSS tokens.

The implementation now matches the framework and web-platform behavior documented by [Next's icon convention](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/app-icons) and the [W3C manifest specification](https://www.w3.org/TR/appmanifest/). The measured maskable monogram remains inside the specified centered safe circle with a radius of 40% of the icon width.

The theme behavior is also coherent: the script-owned unscoped `theme-color` follows the site's resolved theme, manual choices persist, visitors without a saved choice continue following system changes, and client navigation does not duplicate or stale the tag. The no-JavaScript fallback remains light because that is the only theme the CSS can resolve without the bootstrap.

### Security review

No reportable security issue was found. The generator accepts only strict literal colors, verifies the exact downloaded font digest before rendering, and writes to static output paths. Ledger reads are constrained to an explicit repository-relative allowlist. Exported manifest members must remain same-origin and inside the configured deployment base path. Inline theme-script interpolation is limited to validated literals and fixed constants rather than user-controlled content.

### Verification

- Formatting, Biome lint, TypeScript, and `git diff --check`: passed.
- Vitest: 62 files, 506 tests passed.
- `npm run icons`: reproduced all seven binary output hashes from the pinned TTF.
- Production build: passed with 19 generated routes.
- Export verification: 14 HTML pages passed, including the new icon/manifest integrity checks.
- OG integrity check: passed.
- Browser QA: light and dark themes at 1280x720 and 375x812; manual toggle; client navigation; one synchronized `theme-color`; expected icon/manifest links; no horizontal overflow; no console warnings or errors.
- Hosted GitHub Actions: code quality, tests, macOS, Windows, Ubuntu, Node 22, and Node 26 all passed.

### Residual caveats

The previous theme implementation stored inferred system values, so old stored `light`/`dark` values cannot be distinguished from genuine explicit choices. Preserving them avoids erasing a user's real choice, but some returning visitors may need to use the toggle once before future behavior matches their intent. This is a migration limitation, not a current logic defect.

Cloudflare Pages succeeded on the previous head and failed on this head with only “Build failed” and no accessible log or annotation. The current executable changes are generator/test/export-verifier code and all production builds are green. That makes a transient or provider-environment failure more likely, but that remains an inference until the external deployment is rerun or its private log is inspected.

PR #946, opened 2026-07-28
feat(stats): render a real age snapshot and exact build revision

mldangelo:
Stacked on #938.

## What changed

`/stats` used to ship `--.-----------` as the age when JavaScript was unavailable. It also labelled GitHub's repository-wide `pushed_at` value as "Last updated", even though that timestamp does not identify the code that produced the page.

This PR now:

- renders a two-decimal age snapshot on the server, with the date it was taken
- upgrades only that value in the browser while keeping the table server-rendered
- leaves the dated snapshot in place for reduced-motion users and hidden tabs
- reports the exact build commit and UTC build date
- links measured source files and the source tree to the full immutable build SHA
- labels `pushed_at` as the latest repository push and links to repository activity

The first version added a second live timer for build age. Review showed that the static UTC date was clearer and avoided another auto-updating value, so that clock and its abstraction layer were removed.

Build identity is accepted only as a complete `BUILD_SHA` / `BUILD_REPOSITORY` pair or a complete GitHub runner pair. Partial values are rejected instead of being combined with another source or an upstream guess. Local builds omit the commit row.

## Verification

- formatting, Biome, TypeScript, and `actionlint`
- 516 tests
- fork-environment regression test
- production static build: 16 routes
- export verification: 14 pages
- browser QA at 1512px and 320px in light and dark themes
- no-JavaScript output, hydration, reduced-motion behavior, exact commit links, and client bundle boundaries

PR #945, opened 2026-07-28
fix(print): preserve resume links and contact details on paper

mldangelo:
Stacked on #934 (`fix/resume-chronology`). This branch now includes #934 through `d75089b`, so the printed artifact uses the current chronology and role-tier behavior.

## Problem

The screen resume linked employers, schools, and courses, but most of those destinations disappeared on paper. The print header also omitted the location and LinkedIn profile, the screen-only skip link could become a visible print button, and a saved dark theme left dark physical page margins in Chrome's PDF output.

The first implementation exposed every course URL as text. That made the course section much harder to read, so this version uses a narrower policy: print the identifying employer and school URLs, keep course links as PDF annotations, and show the contact destinations as readable link text.

## What changed

- Print the URL after each employer and school name without adding URL blobs to role summaries.
- Add a print-only contact line with location, site, email, GitHub, and LinkedIn. Contact records now have stable IDs, so display-label edits cannot break the lookup.
- Keep each printed contact item intact and separate items with flex gaps instead of punctuation nodes, so an A4 line wrap cannot strand a dot at either edge.
- Hide the skip link and other screen chrome in print.
- Force `color-scheme: light` under `@media print`, including when the saved site theme is dark.
- Let both course-grid tracks shrink with `minmax(0, 1fr)`. This prevents long titles from crossing the A4 column gutter.
- Keep all 13 course links clickable in PDFs without printing their raw URLs.
- Replace stale or unsafe destinations: Arthena now points to its Y Combinator profile, Planetary Resources to a stable retrospective page, Smile ID to `smile.id`, and Skeptical Investments to HTTPS.
- Correct Stanford records that did not match the displayed course numbers: CME 306 is Numerical Solution of Partial Differential Equations, CME 308 is Stochastic Methods in Engineering, and Randomized Algorithms and Probabilistic Analysis is CS 265 / CME 309. CME 302 now links directly to its Stanford Bulletin record.
- Replace brittle assertions about CSS source order and arbitrary URL length with checks for the real selector contract, print import order, A4 shrink behavior, light page canvas, stable retired-company destinations, canonical URLs, and the corrected course records.

## Verification

- `npm run format:check`
- `npm run lint`
- `npm run type-check`
- `npm test`: 60 files, 516 tests
- `npm run build`: 16 static routes
- `npm run verify-export`: 14 pages verified
- Letter PDF: 5 pages, 612 x 792 pt
- A4 PDF: 5 pages, 595.92 x 841.92 pt

[...]

PR #947, opened 2026-07-28
feat: fifteen site improvements, integrated

mldangelo:
Integration branch for the fifteen PRs opened from a survey of every surface of
the site. Per `AGENTS.md`: *"If multiple PRs need to land together, open an
integration branch PR; do not locally merge into main."*

Each constituent PR is reviewable on its own and green in CI. This branch exists
because they were authored in parallel off `main` and collide when combined —
mostly in `AGENTS.md`, where every branch appended to the same Key Patterns
list.

## What it contains

| PR | Change |
|---|---|
| #934 | `fix(resume)` sort the career spine, derive tenure, drop typed "15+ years" |
| #945 | `fix(print)` reveal external URLs, complete the printed contact block |
| #938 | `fix(stats)` count what the page asserts, give every reading provenance |
| #946 | `feat(stats)` age readable with JS off, real deployed commit |
| #941 | `fix(a11y)` WCAG on the amber mark, focus rings, forced-colors |
| #932 | `refactor` delete the dead view transition and the empty forms stub |
| #943 | `fix(writing)` image captions, wide figures, code-fence plates |
| #936 | `fix(about)` the history log's missing years |
| #933 | `fix(a11y)` announce the skills filter, retire the dead colour field |
| #939 | `feat(resume)` serve `/resume.json` |
| #942 | `feat(writing)` a share card per post, digest-gated |
| #940 | `feat(projects)` a register of shipped work |
| #937 | `feat(theme)` ship the control in the HTML, add a system state |
| #935 | `chore(ci)` measure the export and ratchet a size budget |
| #944 | `chore(icons)` retire the orphaned pre-redesign icon set |

## Conflicts resolved

- **`AGENTS.md`** (7 branches) — additive bullets, unioned.
- **`app/resume/page.tsx`** — #934 added a comment explaining the derived
  experience figure; #939 wrapped the title in a header row with the JSON
  affordance. Both kept.
- **`scripts/verify-export.mjs` / its test** — #942 added draft-asset gates,
  #939 added `resume.json` gates. Both kept; the summary line names both.
- **`package.json`** — `icons` and `measure-export` scripts, both kept.
- **`docs/adapting-guide.md`**, **`app/layout.tsx`** — additive.

## One genuine cross-PR failure

Every PR is green alone; together one test failed. `verify-export`'s share-card
test asserted `2 pages OK`, but #939 added a third page to the shared fixture.
Fixed in the integration commit, not papered over. **This is the interaction
per-PR CI structurally cannot catch** — and note that #945 and #946 never ran
the matrix at all, because CI is `pull_request: branches: [main]` and they
target their parents. Both were verified locally (504 and 538 tests).

## Verification

```
npx biome check .   221 files, clean
npx tsc --noEmit    clean
npx vitest run      80 files, 793 tests passed  (main: 57 files, 451 tests)
npm run build       static export OK, 16 routes
npm run verify-export   14 pages OK
npm run measure-export  7 budgets within limits
```

Spot-checked in the real `out/`:

- `/stats` static HTML now carries `36.47` with `as of 2026-07-28` and
  `--readout-width:14ch`; **zero** `--.-----------` placeholders remain.
- Exactly three per-post share cards, one per *published* post. The
  `draft: true` post has none — the leak guard holds.
- `out/resume.json` emitted at 10,643 bytes, file-like route, no trailing slash.
- Amber recomputed independently: `--color-signal-mark` 2.36:1 → **3.51:1** on
  `--color-bg` and 2.16:1 → **3.21:1** on `--color-bg-alt`, both clearing
  SC 1.4.11's 3:1. Text-safe `--color-signal` at 5.48:1 / 5.01:1.

## Two things to look at before merging

1. **The amber is now much darker.** `--color-signal: #9a5600` reads closer to
   dark ochre than amber on cream. It is correct by measurement, but it is a
   visible identity change and the palette is the site's signature — worth your
   eye, not just the contrast checker.
2. **Reverse-chronological sorting moves Arthena below Matroid and Planet**,
   because both started later while running inside Arthena's window. That is the
   standard convention and what was asked for, but it demotes an 8-year
   co-founder role beneath a 9-month and a 7-month one. #934 has a before/after
   table.

## Follow-up this surfaced

The size gate's first real action was to flag repeated inline FontAwesome SVG:
71,459 bytes duplicated across documents, so every new route pays for the icons
again. Ratcheted deliberately here; the architectural fix is its own change.


mldangelo:

mldangelo (commented):
Verdict: changes requested before this integration branch merges to `main`.

The integration is broadly coherent and the fixes added after the initial stack materially improve it. In particular, I verified that the manual theme now updates all emitted `theme-color` values, the new `/resume.json` route exports correctly, and the main routes render without broken images or horizontal overflow at desktop and 390×844. The full quality suite is green. Two concrete defects still need to be resolved, though:

1. The printed résumé still emits the skip link as a large black `SKIP TO CONTENT` button near the bottom of multiple Letter pages. `app/styles/print.css:47-62` hides most screen chrome but omits `.skip-link`; the later integration changes only add `.resume-json-link` and log-marker styling, so the defect reproduced on #945 remains here. This is not cosmetic at the margin—the button is one of the darkest/largest elements on the page and competes with résumé content. Add `.skip-link` to the print hide list, retain a print regression assertion, and inspect the actual PDF again rather than relying only on computed styles.

2. The repository now gives maintainers mutually contradictory factual guidance about `INLINE_YEAR`:
   - `src/lib/logEntry.ts:95-96` says the preposition anchor protects “a Sony Mavica MVC-FD71” from being read as a date.
   - `AGENTS.md:131` repeats that “Mavica MVC-FD71” would start becoming a date without the anchor.
   - `AGENTS.md:132`, added by the follow-up audit, correctly says that example can never reach `YEAR_PATTERN` because it contains no four-digit run and that the old test was ineffective.

The new test is right; the adjacent source comment and older AGENTS rule are wrong. Adding a second bullet that refutes the first leaves the codebase less trustworthy, not more. Rewrite the source comment and consolidate the AGENTS guidance into one accurate explanation: an unprefixed four-digit model such as `Tandy 2000` exercises the year anchor, while “approximately 50 countries” exercises the age anchor. Remove the Mavica claim.

Verification on `0067fa43f74620454e9ea44e5c2673f0eaf7d09f`:

- `npm run format:check`: passed (199 files)
- `npm run lint`: passed (224 files)
- `npm test`: 83 files, 939 tests passed
- `npm run og:check`: all four cards current
- production build with this exact SHA: passed; 16 routes exported, including `/resume.json`
- `npm run verify-export`: 14 pages OK
- `npm run measure-export`: 168 files / 3,932,522 bytes; all seven budgets passed
- browser QA on `/about/`, `/resume/`, `/projects/`, `/writing/`, and `/stats/`; no broken images or horizontal overflow
- mobile QA at 390×844, including opening the navigation menu; no overflow
- manual dark/system/light cycling: computed backgrounds and `theme-color` metadata stayed synchronized
- real Letter-PDF comparison identified the print blocker above
- live GitHub checks are green across test, code quality, Node 22/26 on Ubuntu, Node 26 on macOS/Windows, CI Success, and Cloudflare Pages

This stack does improve the site and most of its added complexity is justified, but the print artifact and contradictory maintenance guidance should be fixed before the protected branch receives it.

mldangelo:
@codex review — the head changed after the earlier review. Please review the current head critically for correctness, integration regressions, accessibility, design, copy, and stale or contradictory guidance.

mldangelo (commented):
This review supersedes [review #4804245450](https://github.com/mldangelo/personal-site/pull/947#pullrequestreview-4804245450) and is anchored to the new head `4a904efddd76f95af8409809f9e73da1776f6945`.

Verdict: changes are still required before this integration branch merges to `main`.

I reviewed the exact `0067fa43..4a904ef` delta (17 files, +569/−98) and then revalidated the resulting integration tree. Several of the seven follow-ups are solid: the icon test now uses URL/posix paths, the invalid view-transition config guard is removed, résumé chronology tests freeze the clock, `Job` receives one shared instant, malformed numeric entities can no longer crash `/resume.json`, and the Stats fallback/profile copy no longer exposes a repository path or manually duplicated refresh date. The export-measurement changes also correctly reject non-object roots and non-integer/negative budget values.

The branch still has the following concrete problems:

1. **The printed résumé still ships the large black skip-link button.** `app/styles/print.css:47-62` still omits `.skip-link`. I rendered this exact head to a five-page Letter PDF; `SKIP TO CONTENT` appears four times in extracted text, and page 1 visibly ends with the large black button. This was a blocker in the superseded review and the new delta does not touch it. Add `.skip-link` to the print chrome suppression and inspect the resulting PDF.

2. **The replacement view-transition rationale is still factually false, and the supposedly dead behavior is reachable.** `app/styles/__tests__/stylesheets.test.ts:137-139` says every internal link uses Next `<Link>`, so no same-origin cross-document navigation occurs. But `src/data/about.ts:39` contains `[Good design](/)`, rendered by the default `markdown-to-jsx` path at `src/components/About/Sections.tsx:185`. The built `/about/` HTML contains a native `<a href="/">Good design</a>`. That is exactly the kind of navigation for which the deleted `@view-transition { navigation: auto }` rule could run. Removing the invalid config-policy test is right; retaining a false “dead code” explanation is not. Either route internal Markdown links through Next, or describe this as a deliberate behavior removal and decide whether that loss is wanted.

3. **An empty budget object still turns the gate off while exiting green.** `scripts/measure-export.mjs:379-417` rejects invalid root/value types, but `{}` or `{ "_policy": [] }` passes. Lines 428-440 then give every metric an undefined limit, and the script reports `0 budget(s) within limits` with status 0. That is the same silent no-gates failure the new root validation says it prevents. Require at least one known metric during normal runs; continue allowing an empty starting point for `--update-budget` if needed.

4. **The newly documented machine-readable command is not machine-readable.** The header advertises `npm run measure-export -- --json -` as raw stdout, and the tests spawn `node` directly. On this head, piping that documented npm command to `jq empty` fails with exit 5 because npm writes its two-line script banner to stdout; the producer then receives `EPIPE`. Direct `node scripts/measure-export.mjs --json -` parses, as does an npm invocation with `--silent`. Document and test the command users are actually told to run.

5. **The proven Mavica/year-guard contradiction remains unchanged.** `src/lib/logEntry.ts:95-96` and `AGENTS.md:131` still claim `Mavica MVC-FD71` exercises the year guard; `AGENTS.md:132` correctly explains that it cannot because it contains no four-digit run. Rewrite the source comment and consolidate the two AGENTS rules (`Tandy 2000` for the year/preposition guard; “approximately 50 countries” for the age guard).

There are also two smaller accuracy/copy cleanups worth making while this is open:

- `src/components/Template/ThemeToggle.tsx:64-68` manually writes “system → light → dark → system” and then says that order is “not a copy” and cannot drift. The sentence is its own counterexample; the executable labels use `nextThemeChoice`, but this prose is manually duplicated. Remove the copied sequence or stop claiming it cannot drift.
- `src/components/Stats/Site.tsx:126-129` still says “no client-side JavaScript shipped,” although this PR deliberately renders the `BuildClock` client leaf. Public labels also still lowercase the GitHub brand (“Stars this repository has on github”).

Verification on the exact new head:

- changed-area suite: 11 files, 176 tests passed
- full Vitest: 83 files, 964 tests passed
- format check, Biome lint, TypeScript, `git diff --check`: passed
- `npm run og:check`: four cards current
- production build with `BUILD_SHA=4a904ef…`: passed; 20 static-generation paths
- `npm run verify-export`: 14 pages OK
- export measurement: 168 files / 3,933,729 bytes; all seven committed budgets passed
- direct `node … --json - | jq`: parsed; documented npm form reproduced the failure above
- desktop and 390×844 browser QA on Stats and Resume: no overflow or broken images; theme choice, computed background, and `theme-color` remained synchronized
- `/resume.json` parsed with 10 roles, canonical metadata, and no misleading `lastModified`
- exact-head Letter PDF reproduced the four skip-link artifacts
- live GitHub checks: tests, code quality, Node 22/26 Ubuntu, Node 26 macOS/Windows, CI Success, and Cloudflare Pages all green

Green CI is credible for the code paths it covers, but it does not resolve the print artifact, the reachable raw-link transition path, the zero-budget configuration, or the contradictory maintenance guidance.

mldangelo:
All five blockers and both cleanups are being addressed. Status:

| # | Item | Where |
|---|---|---|
| 1 | Print skip-link | in flight |
| 2 | Internal Markdown links through Next | in flight |
| 3 | Empty budget gates nothing | #971 |
| 4 | Documented `--json -` not machine-readable | #971 |
| 5 | Mavica/year-guard contradiction | #969 |
| — | ThemeToggle self-contradicting docstring | #967 |
| — | `/stats` "no client JS" claim + GitHub casing | #968 |

**On blocker 4 — you caught my verification error, not just the bug.** I checked `--json -` with `npm run measure-export --silent -- --json -`. The `--silent` hides the two-line npm banner, so it parsed and I reported it fixed. The documented form fails exactly as you describe. Both the docs and a test now use the command a user is actually told to run.

**On blocker 2**, I took your first option rather than rewriting the rationale. Routing internal Markdown links through Next makes the claim true and fixes a real inconsistency — `[Good design](/)` currently triggers a full document load from a client-rendered page.

**On blocker 5**, AGENTS.md contained both the wrong claim and its own correction two lines apart. Consolidating to one bullet: `Tandy 2000` for the year/preposition guard, "approximately 50 countries" for the age guard, with each example verified against the real extractor rather than swapped for another unverified one.

Separately, a re-run of an audit whose verification pass had died on network errors surfaced two more, both now fixed or in flight:

- **`schema.ts` and `Footer.tsx` read `work[0]`** for the site-wide employer — correct only because the newest role happens to be written first, while this branch documented in two places that source order is *not* load-bearing. #970.
- **`theme-color` reverts after the first client-side navigation.** Reproduced in a real browser: React re-creates the meta nodes on every transition and only one survives holding the mutation. It is direction-dependent — device-light + pinned-dark breaks, the reverse survives by tree-order luck, and reversing the declaration array moves the breakage to the other direction. In flight.

I will re-verify the whole tree and update here once the in-flight items land.

mldangelo:

mldangelo (commented):
Fresh exact-head re-review of `70a3dc3011e878547ac35c201ed260be2bcfb586`.

## Verdict: changes required before this integration branch lands

This head is materially better than the prior integration revision. It fixes several real defects: the false view-transition policy guard is gone; internal Markdown links use Next navigation; career ordering now handles the ongoing part-time role; invalid résumé Unicode and fabricated `lastModified` are fixed; legacy article fragment aliases are restored; print suppresses the skip link; and theme-color now follows manual/system choice across client navigation. The integration PR is still useful as the place to reconcile this coupled work.

However, passing 1,052 tests does not make the composed product ready. The current head retains multiple independently reproduced blockers and adds two integration-only regressions.

### Privacy gate is green while knowingly shipping unpublished work

The exact build emits three `verify-export` warnings for `codex-app-overview.webp`, `codex-editor-diff.webp`, and `codex-automations.webp`, all publicly fetchable under `/images/writing/codex-desktop-app-post/`. Commit `6835ecd` detects roughly 116 KB of images referenced only by an unpublished article but deliberately leaves the condition warning-only, so CI exits 0 and deploys them. Move draft-only assets outside `public/` or make this gate fatal; do not call this privacy check clean while it knowingly publishes the files.

### Generated-art reproducibility is still incomplete

Both `scripts/generate-og.mjs` and `scripts/generate-icons.mjs` fetch mutable Google Fonts CSS and font URLs. The OG and icon ledgers hash repository inputs, but not the resolved URLs or downloaded font bytes. The same commit can therefore regenerate different pixels while the freshly written ledger declares them current. Vendor a supported font or version/hash the exact bytes. The post reader's safe-side draft handling improved, but impossible dates still pass its regex-only validation even though the application rejects them.

### Accessibility gates still certify false results

`src/lib/contrast.ts` accepts four/eight-digit colors and discards alpha. Transparent black against white is consequently scored as opaque black—21:1 instead of 1:1—and the tests pin that behavior. Reject alpha or composite against an explicit backdrop. The rationale also still says the amber dot is the only current-role cue despite visible `PRESENT` text.

`PostContent.tsx` still gives every code fence `tabIndex=0` and `role=region`, even when it does not overflow; repeated Markdown fences receive the same region label. Repeated normalized headings can still receive duplicate canonical IDs because the renderer calls the stateless slugger directly even though a unique-ID helper exists. Legacy aliases are fixed, but canonical uniqueness and landmark discipline are not.

### Public counts and labels remain misleading

`countLockedPackages()` excludes all optional/devOptional entries, while `/stats` labels 54 as “Installed to run in production.” The exact lock represents 15 compatible production optionals plus a production-reachable devOptional, producing 70 for the target platform; more fundamentally, this statically exported site installs no packages at runtime. Count the declared concept accurately and name it accordingly.

The new deployment row is useful, but its fork fallback is not self-sufficient: `deployedCommit()` can accept a fork's `GITHUB_SHA`, while `src/data/stats/site.ts` hardcodes `mldangelo/personal-site` for the link. An unadapted fork can therefore point its own SHA at an upstream 404. GitHub exposes the current `owner/repo` through `github.repository`/`GITHUB_REPOSITORY`; plumb that identity with the SHA or narrow the fallback contract and guidance.

### Résumé and history prose still claim more precision than the data supports

The career spine ordering is fixed, but the headline still presents elapsed calendar span from June 2011—including gaps—as “years building products,” and each terse duration (for example, `4 mo`) is appended without an announced label. Name this as elapsed span or calculate the intended active-time concept, and expose the duration meaning accessibly.

The About transformer still deletes “summer” from “In the summer of 1996,” manufactures exact age/year pairs from only a birth year despite the documented birthday ambiguity, and leaves the source sequence nonchronological (`2006 → 2004–2007 → 2008`). That is not one coherent temporal system.

### Export and project facts need correction

The `/resume` affordance remains a native `href="/resume.json"`; Next does not prefix native anchors with a repository `basePath`, and the verifier masks this by testing a hand-authored already-prefixed fixture rather than the rendered page.

Project data still has several concrete problems:

- `http://www.spacepotato.org` does not resolve.
- Cat Detector is dated May 15, 2015 and tagged TensorFlow, while TensorFlow's official history says it was open-sourced on November 9, 2015. Correct the date/tag or explain what the date means: https://blog.tensorflow.org/2022/10/building-the-future-of-tensorflow.html
- Promptfoo is simultaneously modeled as an ongoing project and an ended CTO employment interval, leaving readers unable to tell which dimension the date/status represents. Promptfoo's current About page does support “CTO & Co-founder”; the title itself is not the problem.
- The blanket sourcing guarantee remains stronger than the tests, and the ordering test collapses ongoing entries to the same `Infinity` key.

A live correction: `arthena.com` currently resolves to the legitimate `arthena.co` site and states the Masterworks acquisition. Do **not** retain the earlier “currently hijacked” claim as a present blocker; at most it was intermittent/previously observed.

### Integration-only maintainer contradiction

`AGENTS.md` still mandates a media-scoped `theme-color` pair and says bootstrap/ThemeToggle repoint every tag. This branch intentionally replaced that architecture with one unscoped script-owned tag plus a light `<noscript>` fallback. Update the guidance to the actual invariant and navigation model; the branch should not teach future agents to undo its own fix.

## Fresh exact-head verification

- 344 targeted tests passed
- full suite: 84 files / 1,052 tests passed
- Biome, Prettier, TypeScript, `git diff --check`, and `og:check` passed
- production build generated 20 paths
- export budgets passed
- `verify-export` exited 0 but printed the three publicly fetchable draft-image warnings above
- component branches also received responsive/browser or PDF QA appropriate to their surfaces
- worktree remained clean

The integration work is necessary, and many audit fixes are good. It should land only after the remaining factual, privacy, reproducibility, accessibility, base-path, and maintainer-guidance issues above are reconciled on this exact composed branch.

mldangelo:

mldangelo (commented):
Final exact-head re-review at 3e1681e.

Disposition: no remaining blocker. I re-audited the integrated result rather than assuming the standalone approvals composed cleanly. The conflict resolutions preserve the stronger three-state theme architecture, fail-closed export policy, shared publication validator, honest chronology/project semantics, portable generated assets, and print safeguards.

This pass found one integration-only defect that the earlier branch checks missed: a real-image unit test still referenced a draft article asset that the new draft-export gate correctly deleted. The test now uses a published JPEG while the synthetic cases continue to cover all WebP encodings. The full suite is green at 84 files / 1,072 tests. Format checking (203 files), lint (228), type checking, four-card Open Graph verification, production build, 14-page export verification, and all seven export budgets pass; total export is 3,816,153 bytes.

Browser QA covered /, /about/, /resume/, /projects/, and /writing/ at desktop and 320px with no document overflow or broken assets; live system-theme behavior, theme-color persistence across navigation, mobile modal focus/scroll lock, skills filters, overflow-only code focus, and deep-link offsets all behave correctly. The five-page résumé PDF contains all 27 skills with no clipping, overlap, repeated chrome, or glyph defect. GitHub’s Linux, macOS, Windows, test, quality, aggregate, and Cloudflare checks are green.

PR #948, opened 2026-07-28
fix(resume): drop a misleading lastModified and correct a false comment

mldangelo:
Two low-severity defects in `src/lib/resumeJson.ts`, both confirmed on inspection.

## 1. `meta.lastModified` meant something narrower than its name

`lastModified()` collected only `startDate`/`endDate` from `src/data/resume/work.ts` and emitted the maximum. So the field that named itself "last modified" was the field least able to notice a modification: editing `profile.json` (name, role, email, city), `contact.ts` (all six profile URLs), `degrees.ts`, `courses.ts`, or every entry in `skills.ts` left the timestamp untouched.

**The choice, and why.** The field is dropped rather than widened.

The alternative offered was to derive it from the newest date across all dated data files. I checked what that actually yields: `courses.ts`, `skills.ts`, and `contact.ts` carry no dates at all, `degrees.ts` carries years (2016, 2012), and `profile.json` carries a birth date. The maximum across every dated file is therefore *the same work-history date the old implementation already published*. Widening the search would broaden the claim without broadening what the claim can see — a wider lie for identical output. It also would not have fixed the reported symptom: edit every skill, and the timestamp still does not move.

Reading the clock is out: it rewrites the artifact on every rebuild, churns it in git, defeats caching, and cannot be pinned by a test — the discipline `feed.xml` already applies to `lastBuildDate`. Reading git or file mtimes is out too: `actions/checkout` clones at depth 1, so `git log -1 -- src/data` returns nothing on most builds, and a fresh clone's mtimes are the checkout time, i.e. the build clock wearing a hat. Hand-maintaining a date in a data file is the failure mode AGENTS.md already names — a figure typed by hand with nothing checking it.

`meta.lastModified` is optional in JSON Resume, so no field ships. `work[0].startDate` already states the recency of the work history under a name that is true, and `meta.canonical` is unchanged. `scripts/verify-export.mjs` checks the root key set and `meta.canonical`, neither of which is affected.

## 2. The `toPlainText` doc comment was false

It claimed stripping inline anchors cost nothing because "every employer's own `url` is already a field on the work entry." It is not. The six hrefs are third-party:

| href | entry `url` |
| --- | --- |
| `openai.com/index/codex-security-now-in-research-preview/` | `https://openai.com` |
| `anthemis.com`, `foundationcapital.com`, `ycombinator.com` | `https://arthena.com` |
| `nea.com`, `accel.com` | `https://matroid.com` |

None appears anywhere in the artifact. The **behaviour** is correct — JSON Resume prose is plain text, and the schema has nowhere to hang a link inside a summary — so nothing about the output changes. Only the justification is rewritten, to say what is true: the anchor text is preserved, the href is dropped, and these particular URLs do not survive.

**On preserving them:** I judge none worth keeping. Doing so would need an invented `meta` key or a fake `projects`/`references` entry, and the one thing that actually fails JSON Resume validation is a non-schema property. The Codex Security announcement is the only one a reader might want, and it is reachable in one click from `https://openai.com`, which the entry already carries.

## Tests

The removed `lastModified` assertion is replaced by the property it existed to protect, generalized rather than deleted: `serializeJsonResume()` must return byte-identical output after `vi.setSystemTime` moves the clock five years. That covers any future field, not just the one that was there.

Two new tests convert the corrected comment into an enforced property, by extracting every `<a href>` from the source prose and asserting that (a) each anchor's text reaches the built prose, (b) no anchor's href appears anywhere in the serialized document, and (c) no anchor's href equals any entry's `url` — the assertion that directly refutes the sentence this PR removes. If a summary gains or loses an anchor, the count assertion fails and the next author has to re-read the comment.

## Verification

```
npm run format
npx biome check .   # exit 0
npx tsc --noEmit    # exit 0
npx vitest run      # 80 files, 797 passed (793 on base, +4 here)
```

Files touched: `src/lib/resumeJson.ts`, `src/lib/__tests__/resumeJson.test.ts`. `buildWork()`'s `sortPositions` change is untouched.

### DIFF
--- a/src/lib/__tests__/resumeJson.test.ts
+++ b/src/lib/__tests__/resumeJson.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 
 import contact from '@/data/contact';
 import profile from '@/data/profile.json';
@@ -39,6 +39,29 @@ const ISO8601 =
   /^([1-2][0-9]{3}-[0-1][0-9]-[0-3][0-9]|[1-2][0-9]{3}-[0-1][0-9]|[1-2][0-9]{3})$/;
 
 const resume = buildJsonResume();
+
+/**
+ * Every inline `<a href>` in the source résumé prose. The doc comment on
+ * `toPlainText` claims the text of each one survives into the artifact and the
+ * href does not; these are what hold it to that.
+ */
+const sourceAnchors = work
+  .flatMap((position) => [
+    position.summary ?? '',
+    ...(position.highlights ?? []),
+  ])
+  .flatMap((source) => [
+    ...source.matchAll(/<a\s[^>]*href='([^']+)'[^>]*>([^<]+)<\/a>/g),
+  ])
+  .map((match) => ({ href: match[1], text: match[2] }));
+
+/** The plain-text prose of the built document, as one string. */
+const resumeProse = resume.work
+  .flatMap((position) => [
+    position.summary ?? '',
+    ...(position.highlights ?? []),
+  ])
+  .join('\n');
 
 describe('toPlainText', () => {
   it('keeps anchor text and drops the markup', () => {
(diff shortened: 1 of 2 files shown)


mldangelo:

mldangelo (commented):
Verdict: the code change is sound and was reasonable to merge into the integration branch; two review nits should still be recorded.

I checked the declared JSON Resume v1.0.0 schema itself: `meta.lastModified` is optional, so omitting it is schema-valid. More importantly, omission is more honest than publishing the newest work-history date as though it represented every data source. Correcting the `toPlainText` explanation is also worthwhile—the old comment made a factual claim the data directly disproved.

Two details are overstated:

- The PR rationale says a clock-derived `lastModified` would “churn it in git.” The generated artifact is `out/resume.json`, and `out/` is gitignored, so it would not churn committed history. It would make every build byte-different, undermine reproducibility/caching, and attach a misleading content-freshness meaning to build time; those are already sufficient and accurate reasons to reject it. Please avoid the git claim in any durable documentation or follow-up summary.
- `sourceAnchors` only recognizes single-quoted, flat HTML anchors (`href='…'` with plain text). A newly added double-quoted anchor or nested markup can evade extraction while the hard-coded count remains six, so the PR’s claim that any gained anchor necessarily fails is broader than the test. The current six anchors are covered, and this is not a blocker, but either make extraction match the forms `toPlainText` accepts or narrow the comment/test claim.

Verification on the exact merged head `1c38b3bb3c173cea5c0ca24b3685c9c59b2c157c`:

- focused JSON Resume route/library tests: 2 files, 26 tests passed
- the containing integration branch’s full 939-test suite, production build, export verification, and `/resume.json` browser/export checks also passed

Net: necessary accuracy fix, valid schema decision, no regression found.

PR #955, opened 2026-07-28
feat: refresh resume skills

mldangelo:
## Summary

Stacked on `integration/site-improvements` (#947), not on `main` — it changes the same data, components, CSS, tests and AGENTS bullets as #933, and the two needed one intentional resolution rather than a race.

- replace the alphabetized, multi-category skills list with a deliberate 2026 taxonomy across agent systems, AI security and evals, ML systems, and software infrastructure
- add current areas such as agent orchestration, MCP tool use, context engineering, agent telemetry, evaluation infrastructure and graders, prompt-injection defense, and sandboxed execution
- remove public proficiency entirely: the 1–5 `competency` score, the `core`/`working`/`familiar` tier derived from it, and the `featured` flag that would have replaced it. Authored order is the only hierarchy
- keep #933's live region, its removal of the dead per-category `color`, and its accessible-name rule
- keep every group mounted behind `hidden` so `print.css` restores it, and suppress the `.skip-link` that was printing on four of five pages

## Why

The previous section understated current AI-agent and software-security work, repeated skills across categories, and exposed subjective 1–5 ratings. The new section is calibrated to senior technical scope without publishing a job level or a self-assessed proficiency of any kind.

## Validation

- `npm run format:check`, `npx biome check .` (224 files), `npx tsc --noEmit` — clean
- `npx vitest run` — 83 files, 954 tests, 0 failing (base `integration/site-improvements`: 83 files, 964 tests; the 10-test delta is entirely the removed `tierFor` / tier-class / competency-range / alphabetical-sort assertions, offset by 3 new guards)
- `npm run build` — static export, 20 routes
- `npm run verify-export` — 14 pages OK · `npm run measure-export` — 7 budgets within limits · `npm run og:check` — pass
- inspected `out/resume/index.html`: 4 groups, none `hidden`, 27 tags each rendered exactly once; no `--tag-color`, tier class, `aria-label` or `title` residue; live region reads `Showing all 27 skills.`
- inspected the shipped CSS: `.skip-link` is inside the `@media print` `display:none!important` rule; `.skill-tag--*`, `.skill-tag-tier`, `.skill-tag-name` and `.skill-tag::before` are all gone
- inspected `out/resume.json`: 27 keywords in page order across 4 groups, no `level`
- **print**: the specific defect from the review — the accent-filled "SKIP TO CONTENT" box at the foot of pages 1–4 — is fixed and pinned by a test that fails when the suppression is removed (verified by removing it). I have **not** re-rendered a Letter PDF myself, so I am not restating that the whole print artifact is clean; the earlier claim to that effect was wrong and is withdrawn.

## Not done

- `featured` was cut entirely rather than trimmed to 6–8. Choosing a "signature" subset is the same self-assessment this PR removes, with two levels instead of five.
- "Agent Harnesses" is now "Agent Orchestration". Restore the stronger term if there is hands-on ownership of an agent loop that the public record does not show.


mldangelo:

src/components/Resume/Skills.tsx
```diff
@@ -50,19 +50,15 @@ export default function Skills({ skills, categories }: SkillsProps) {
...
-      return a.title.localeCompare(b.title);
-    });
-
     return categories
       .map((category) => ({
         category,
-        skills: sortedSkills.filter((skill) =>
-          skill.category.includes(category.name),
-        ),
+        skills: skills.filter((skill) => skill.category === category.name),
```
mldangelo, line 62: This grouping rewrite is also where the active accessibility work needs to be reconciled. Pressing a filter hides most of the section, but this head provides no accessible result announcement: `aria-pressed` reports control state, not that (for example) only 6 of 27 skills remain visible. PR #933 already adds a computed `role="status"` / `aria-live="polite"` result and tests. Please rebase/reconcile and preserve that behavior here rather than letting this taxonomy rewrite silently discard it.

src/data/resume/skills.ts
```diff
@@ -1,141 +1,166 @@
...
-    category: ['Languages', 'ML Engineering'],
+    title: 'Coding Agents',
+    category: 'Agent Systems',
+    featured: true,
   },
   {
-    title: 'TypeScript',
-    competency: 5,
-    category: ['Languages', 'Web Development'],
+    title: 'Agent Harnesses & Orchestration',
```
mldangelo: Please verify this against direct hands-on ownership. “Harness” has a specific, stronger meaning than “agent workflow”: OpenAI describes the Codex harness as the core agent loop and execution logic (https://openai.com/index/unrolling-the-codex-agent-loop/). The public résumé and writing clearly support using/orchestrating parallel coding agents, but currently only say “a few small PRs” were shipped into Codex. Unless there is direct work evidence behind the stronger claim, “Agent Orchestration” or “Coding-Agent Workflows” would be more precise and defensible.

src/components/Resume/Skills/SkillTag.tsx
```diff
@@ -1,39 +1,28 @@
...
-  )?.color;
-
-  // Size based on competency (5 = large, 4 = medium, 3 = small)
-  const sizeClass =
-    competency >= 5
-      ? 'skill-tag--lg'
-      : competency >= 4
-        ? 'skill-tag--md'
-        : 'skill-tag--sm';
+  const categoryColor = categories.find((cat) => cat.name === category)?.color;
```
mldangelo: Every category is assigned the same `var(--color-accent)`, so this lookup does not encode any category distinction. It leaves a `Category.color` field, a `categories` prop on every tag, an inline CSS variable, and a unit test for behavior that is visually constant. Remove the color plumbing and let the tick use the accent token directly; PR #933 independently reached the same simplification, and keeping this dead path creates avoidable merge conflict.

mldangelo (commented):

The direction is worthwhile and materially improves the résumé: removing public 1–5 self-ratings is more credible, the four-category taxonomy is much more current than the old framework/library inventory, authored ordering is clearer, and single-category display removes duplicate tags. The implementation is also mechanically healthy in isolation.

There are four things I would address before this lands:

1. **Reconcile the active skills-accessibility work rather than landing a competing implementation.** PR #933 (also carried by the integration PR #947) changes the same data, components, CSS, tests, and AGENTS guidance. A non-mutating merge simulation between #933 at `79b7152` and this head produced content conflicts in seven files: `AGENTS.md`, `tags.css`, `SkillTag.tsx`, both component test files, the data test, and `skills.ts`. `Skills.tsx` happens to auto-merge and retains #933's live result announcement, but the rest requires an intentional resolution. Please rebase/stack this after the integration branch and preserve the useful accessibility/status work while deliberately superseding the old proficiency-tier presentation.

2. **Calibrate the résumé claims to direct evidence.** “Coding Agents,” multi-agent workflows, MCP-assisted QA, evals, red teaming, static analysis, and production ML all have public support in the résumé/writing. “Agent Harnesses,” however, is a stronger implementation/ownership claim than “using or orchestrating coding agents”; OpenAI's own terminology defines the harness as the core agent loop and execution logic (https://openai.com/index/unrolling-the-codex-agent-loop/). The public copy currently says only that the author has shipped a few small Codex PRs. Keep “harnesses” only if direct hands-on ownership supports it; otherwise “Agent Orchestration” or “Coding-Agent Workflows” is more defensible.

3. **Make the visual hierarchy genuinely selective.** `featured` is set on 16 of 27 skills (59%). When most items receive the special treatment, it stops identifying “signature areas” and instead makes the remaining 11 look weak. At desktop and 320px the weight difference is legible but subtle; I would reduce the featured set to a small, defensible core (roughly 6–8) or remove the second hierarchy entirely and let authored order do the work.

4. **Correct the print-QA result.** I printed the production export to a US Letter PDF and rendered all five pages. The skills fit cleanly on page 4 with no clipping, but a large black “SKIP TO CONTENT” control appears at the bottom of pages 1–4. This is pre-existing rather than introduced here, but it directly contradicts the PR body's claim that print behavior was checked. Since this PR changes the printed résumé, add `.skip-link` to the print chrome suppression (and ideally a regression assertion), or narrow the validation claim rather than calling the print artifact clean.

Verification on the exact head:

- `npm run format:check`, lint, type-check: pass
- Vitest: 57 files / 442 tests pass
- `npm run og:check`: pass
- production static build: pass (16 routes)
- `npm run verify-export`: 14 pages pass
- browser QA at 1280×720 and 320×900: no horizontal overflow; all 27 skills render once; filters wrap without clipping
- print QA: five Letter pages; skill layout is clean, but the skip-link defect above is visible on four pages
- terminology research: the overall categories match current official language around coding agents, multi-agent workflows, MCP, evals/graders, sandboxing, and agent telemetry; the “harness” claim is the one that needs evidence calibration

Because this is our own PR, GitHub does not allow a formal request-changes review; this COMMENT review is an explicit changes-needed verdict.

mldangelo:
All four items addressed. Force-pushed a rebase, so the history moved: `7d0fdf3` (parent `e638bce`, i.e. `main`) → `63c0495` (parent `4a904ef`, the head of `integration/site-improvements`). One commit, same as before; nothing else on the branch. PR base retargeted to `integration/site-improvements` so the diff is the 13 files this change actually touches rather than all of #947.

### 1. Reconciled with #933, not landed beside it

All seven conflicts you predicted appeared, in exactly those files, and were resolved by hand. `Skills.tsx` auto-merged as you said.

**Kept from #933/#947**
- the polite `role="status"` region announcing the *result* (`Showing 2 of 5 skills in Agent Systems.`), counted from the rendered groups, still `.sr-only`
- the removal of the per-category `color` field that handed all four categories the same `var(--color-accent)`
- the accessible-name rule: nothing here relies on an ARIA name sitting on a `generic` element

**Superseded**
- the `core`/`working`/`familiar` tier and the 1–5 score behind it

### The text tier: it goes

Its only input was `competency`, which this PR deletes, so it had nothing left to express. Re-deriving it from `featured` would have been the same self-rating with two levels instead of five. The pipeline is removed end to end — `tierFor`, `SkillTier`, `MAX_COMPETENCY`, `.skill-tag-tier`, `.skill-tag--core/--working/--familiar`, the sr-only `proficiency:` prefix, and the dead `.skill-tag::before` print override left behind when #933 deleted the tick. Nothing half-removed; `grep` for any of it in `out/` returns nothing.

`SkillTag` is now one line: `<span className="skill-tag">{title}</span>`. There is no accessible name to get wrong and no state encoded in colour.

### AGENTS.md

Resolved to one set of rules, no stale bullet beside its replacement. Kept "An accessible name needs a role that accepts one" and "State is not outcome" verbatim — both are general lessons that outlive the tier. Replaced "Competency is text, not colour" (it documents a presentation that no longer exists) with a bullet that records all four removed encodings and forbids a fifth. Added two: the `String.includes`-on-a-scalar trap, and the fixed-position/print rule.

### 2. "Agent Harnesses" → "Agent Orchestration"

Applied your fallback. This is a claim about your own work and I can't settle it — if there is hands-on ownership of an agent loop that the public record doesn't show, restore it in `src/data/resume/skills.ts`; it is one string.

### 3. `featured` removed entirely

Took the cleaner option. The argument that decided it: picking a defensible 6–8 is itself a self-assessment, which is precisely what this PR is removing — it would just have two levels instead of five. Order already carries it. `Agent Systems` leads the four groups, and `Coding Agents`, `AI for Software Security`, `Production ML Systems` and `Python` lead their groups. `.skill-tag` now has exactly one weight, and a test asserts every tag renders one class and no inline style.

### 4. Print

**Your expectation that I would inherit the fix is wrong** — I did not. `.skip-link` is not in the print chrome suppression at `origin/integration/site-improvements` (`4a904ef`), it is not on `main`, and no open PR carries it. So I fixed it here, since this PR changes the printed résumé:

- `.skip-link` added to the `display: none !important` chrome block, with a comment explaining that paged media repeats a fixed element and `transform: translateY(-200%)` is all that was hiding it
- a guard in `app/styles/__tests__/stylesheets.test.ts` pins the **class**, not the instance: every `position: fixed` selector in the stylesheet graph must be named in a `display: none` rule in `print.css`. Verified it can fail — removing `.skip-link` gives `expected [ '.skip-link' ] to deeply equal []`
- confirmed in the shipped CSS, not just the source: `.skip-link` sits inside the `@media print` `display:none!important` rule in `out/_next/static/css/*.css`
- if a dedicated print-skip-link change is in flight elsewhere, this hunk is a duplicate of one selector and should be dropped in resolution

On the PR body: I narrowed the print claim rather than restating it. I have not re-rendered a Letter PDF, so the body no longer says the print artifact is clean — only that the specific defect is fixed and regression-tested.

### One thing your merge simulation could not see

It ran against #933 alone; the integration branch also carries #939. `buildSkills` in `src/lib/resumeJson.ts` sorted keywords by `competency` — a field that no longer exists — and filtered with `skill.category.includes(category.name)`, written when `category` was an array. Against a single string that is **substring** matching, and it is correct only because no category name happens to contain another. Now `===`, keywords in authored order so the artifact and the page agree, and `src/data/__tests__/skills.test.ts` pins the no-substring property so the next category name can't quietly break it.

### Verification

```
npm run format:check          clean
npx biome check .             224 files, clean
npx tsc --noEmit              clean
npx vitest run                83 files, 954 tests passed
npm run build                 static export, 20 routes
npm run verify-export         14 pages OK
npm run measure-export        7 budgets within limits
npm run og:check              pass
```

Base `integration/site-improvements` measures 83 files / 964 tests. The −10 is accounted for exactly: `SkillTag.test.tsx` 10→3, `Skills.test.tsx` 18→17, `skills.test.ts` 12→7 (the old twelve were competency ranges, array-category shape and alphabetical-sort assertions, all superseded), `resumeJson.test.ts` +1, `stylesheets.test.ts` +2. No file lost coverage of anything still true.

Checked in the real `out/`, per the print constraint: **4 skill groups, none `hidden`, 27 tags each rendered exactly once.** `/resume.json` carries the same 27 in page order with no `level`.

### Deliberately not done

- did not trim `featured` to 6–8 (see 3)
- did not re-render the print PDF, so no whole-artifact print claim is made
- did not touch the taxonomy content itself beyond the one rename — you said the four categories match current official language

